PR #41, opened 2015-02-09
Fixed windows build.

ocefpaf:
@rsignell-usgs we forgot the `bld.bat`.  With this PW we can build any OWSLib version and forget about OWSLib issues here https://github.com/ocefpaf/utilities/issues/1.


rsignell-usgs:
Doh!   I'm glad we didn't report that one as a bug.  :flushed:

PR #43, opened 2015-02-12
Create `.travis.yml`.

ocefpaf:
@rsignell-usgs We need to activate travis for this repository (https://travis-ci.org/profile/ioos) to see if this works.  I do not have admin rights to do it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,12 +25,10 @@ install:
     - python bootstrap-obvious-ci-and-miniconda.py ~/miniconda x64 2 --without-obvci && source ~/miniconda/bin/activate root
     # TODO: Change rsignell to ioos.
     - conda config --set show_channel_urls True && conda config --add channels http://conda.binstar.org/rsignell/channel/main
-    # I am going yo keep this for now.  We probably need something from his latest obvious-ci version.
+    # I am going to keep this for now.  We probably need something from his latest obvious-ci version.
     - conda install -c http://conda.binstar.org/pelson/channel/development --yes --quiet obvious-ci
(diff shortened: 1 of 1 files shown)


ocefpaf:
BTW **DO NOT MERGE!**  It is likely that this will fail.  Lets use this PR to keep updating/fixing the yaml file until we get it right and only then merge.

rsignell-usgs:
Like this, right?
2-12-2015 2-06-47 pm

ocefpaf:
Yes, and now travis-ci is working its magic...

rsignell-usgs:
oh no!
OSError: liblas_c.so: cannot open shared object file: No such file or directory

ocefpaf:
... and we have out first error!

ocefpaf:
libLAS (python) depends on libLAS (C and C++) the conda package for the C/C++ is missing.  You probably can do a local built because you have it there (I just checked that, like travis, I cannot.)

PS:  Who gives the library and the wrapper with the same name?

ocefpaf:
I found that "https://github.com/grantbrown/laspy" an alternative to libLAS (python wrapper).  It seems to be better maintained (shows recent activity in Nov 2014) and it is better named.  LibLAS latest version is from 2011:

https://pypi.python.org/pypi/libLAS

@rsignell-usgs Question:

1. Rename libLAS (python wrapper) to make room for libLAS (C/C++)?
2. Rename libLAS (C/C++) to something else and keep libLAS (python wrapper) ?
3.  Delete libLAS (python wrapper) add libLAS (C/C++)  and laspy?

PR #46, opened 2015-02-13
Updated rasterio to 0.18 and use PyPI source.

ocefpaf:
I found that the cython error we got were due to the GitHub tagged version.  I believe there are some left over cythonized sourced there.  Once I moved to a clean source (PyPI) things worked out smoothly for all rasterio version.

Here I also updated it to the latest version (0.18).


ocefpaf:
rasterio build is finally OK!  now `rdflib` is the issue...

PR #48, opened 2015-02-13
Update ulmo.

ocefpaf:
Updated ulmo.  Somehow the old version was trying to download (setuptools download is not allowed during a conda build) a different pandas version.  This one seems to be OK locally.  Now, lets wait for travis...

### DIFF
--- a/ulmo/meta.yaml
+++ b/ulmo/meta.yaml
@@ -1,28 +1,13 @@
 package:
   name: ulmo
-  version: !!str 0.7.4
+  version: !!str 0.7.5
 
 source:
   git_url: https://github.com/ulmo-dev/ulmo.git
-  git_tag: v0.7.4
-#  patches:
-   # List any patch files here
-   # - fix.patch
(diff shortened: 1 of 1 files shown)


ocefpaf:
I like you confidence in my PRs!  You did not even waited for travis :wink:

ocefpaf:
And surprisingly we have the first fully successful build!  The next step is to set the TOKEN up.

rsignell-usgs:
Yippee!

ocefpaf:
A good side effect from using travis to build everything was to find all the "stale" builds we had in the repository.  Hopefully we will be able to do the same from windows.

emiliom:
Very cool stuff the two of you are doing! With this, udunits, rdflib, etc.

ocefpaf:
Thanks @emiliom. Soon we will be testing an automatic build/upload using travis-ci.  When that is in place we will need help to test these packages...  Stay tuned.

Issue #64, opened 2015-02-16
Enable windows build via AppVeyor


ocefpaf:
- [X] add `appveyor.yml`
- [ ] start to :sob:

ocefpaf:
@rsignell-usgs [appveyor](http://www.appveyor.com/) does not open its log files like travis-ci.  Should we use my account or should we create an IOOS account and share the password with https://github.com/ioos/conda-recipes contributors?

rsignell-usgs:
I like the ioos account idea.   Can't see much risk of shared account here!  ;-)

ocefpaf:
OK.  Since you are one of the IOOS GitHub admin can you do this and then you share the password with me?

rsignell-usgs:
@ocefpaf , hmmm, looking into this, we need either an e-mail or a github developer account to create an appveyor account, but we have only an IOOS github organization, not an IOOS github developer account.   And I'm not sure we should create one either.   

We could still create a new appveyor account without associating with github, but then we have to still use some e-mail address to sign up.    We could create an IOOS gmail account or something, but perhaps we should just back off and use your (or my) appveyor account for now.   Unless I'm missing something.   I wonder what @pelson and the scitools gang did for an appveyor account.

ocefpaf:
OK.  To move this forward I will use my account.  But in the long run I believe that someone affiliated to IOOS should take over.

rsignell-usgs:
The windows builds certainly are going to be challenging.   Do you think we should pursue the MinGW approach or try to get stuff going with the CMAKE w/ Native Windows Compiler approach?   I wonder how Golhke builds his [windows binaries for python packages](http://www.lfd.uci.edu/~gohlke/pythonlibs/).  

On a related note, I'm not sure a Native Windows Fortran compiler is available on Appveyor, but of course we would use gfortran with MinGW.

rsignell-usgs:
Some interesting links here?
https://github.com/JuliaLang/julia/blob/master/appveyor.yml
https://github.com/JuliaLang/julia/blob/master/README.windows.md

ocefpaf:
> Do you think we should pursue the MinGW approach or try to get stuff going with the CMAKE w/ Native Windows Compiler approach?

No idea.  My experience here is zero.  My guess is that CMAKE is more "AppVeyor friendly."

> I wonder how Golhke builds his windows binaries for python packages?

You are not alone there...

[...]

PR #65, opened 2015-02-16
Added `appveyor.yml`.

ocefpaf:
@rsignell-usgs No build will take place until we sort  #64 out.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,42 @@
+environment:
+
+  CONDA_INSTALL_LOCN: "C:\\conda"
+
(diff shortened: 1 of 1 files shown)


pelson:
Pleasantly short PR :wink:.
For what I is worth, I found appveyor a little slow to spin up, so I did all of my building locally and then just pushed the recipes once they actually worked.

ocefpaf:
Thanks @pelson.  That is basically a copy-and-paste from your yaml file :grimacing: 

AppVeyor slowness is already biting me.  But I do not have a windows machine at my disposal (yet).

PR #73, opened 2015-02-18
Fix `sci-wms-deps` windows build+rename package.

ocefpaf:
@rsignell-usgs the package was renamed from `sci-wms` to `sci-wms-deps` to be more honest with the user :stuck_out_tongue_winking_eye:


rsignell-usgs:
good deal.

PR #75, opened 2015-02-18
Fix `rdflib` windows build.

ocefpaf:
Using `pip` instead of `wget` the circular dependency work around it OS agnostic (and more robust).

### DIFF
--- a/rdflib/bld.bat
+++ b/rdflib/bld.bat
@@ -1,8 +1,3 @@
+pip install SPARQLWrapper
 "%PYTHON%" setup.py install
 if errorlevel 1 exit 1
-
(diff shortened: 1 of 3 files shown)


rsignell-usgs:
Why are we using `pip` here?

ocefpaf:
That is the circular dependency work-around.  I "inject" `SPARQLWrapper` otherwise `rdflib` will not build.  The previous version used `wget` to download the egg binary.  I changed to pip so it can be easily ported to windows.

PS: That is a trick used in the build, but both `SPARQLWrapper` and `rdflib` are real conda packages.

rsignell-usgs:
okay, I get it now.

PR #77, opened 2015-02-18
Adding psycopg2 (Cheating: using pip).

ocefpaf:
@rsignell-usgs This will download a binary from PyPI using pip and create a conda package from it.  I remember reading about this in the conda e-mail list somewhere...

The psycopg2 package works only on win64 and linux (no win32).  With that we can build `pycsw` for [win64](https://binstar.org/ioos/pycsw).  However, both packages are not tested.

### DIFF
--- a/psycopg2/bld.bat
+++ b/psycopg2/bld.bat
@@ -1,0 +1,1 @@
+pip install psycopg2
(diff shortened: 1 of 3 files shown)


rsignell-usgs:
Cheaters never prosper?  Bah!

ocefpaf:
I know.  As a purist I did not like that approach... But look, pycsw is on windows!  (This statement is untested and the general surgeon says: using windows is bad for your health.)

PR #80, opened 2015-02-20
First step to fix windows builds.

ocefpaf:
Removed the following windows build to allow AppVeyor to complete.

rtree/bld.bat
octant/bld.bat
pytools/bld.bat
rasterio/bld.bat
udunitspy/bld.bat
mysql-python/bld.bat
compliance-checker/bld.bat

I will revisit every package again at a later time to fix their windows build individually.

PR #82, opened 2015-02-20
Make packages consistent by using the same numpy.

ocefpaf:
- Removed numpy 1.8 from the build matrix
- Removed `cdat-lite` and `cfchecker` that depends on on numpy 1.8

@rsignell-usgs this PR removes two packages that, we assume, are not widely used by the community.  The advantages is that the remaining package will be "numpy" consistent.  One side effect is to get the latest netcdf4-python with the latest libnetcdf.


ocefpaf:
The failures here will go away with #80.

pelson:
P.S. https://github.com/cedadev/cf-checker/issues/2

ocefpaf:
Thanks @pelson!  That is good to know.  I am a user of cf-checker.  However, I IOOS guys are focusing their effort into their own tool.

https://github.com/ioos/compliance-checker

So the decision to leave those packages out was not only numpy, but also the low user demand for them.

PR #83, opened 2015-02-20
Added iris for windows.

ocefpaf:
This PR adds iris build for windows (`bld.bat`) and updates iris to the latest release.

The update is because version 1.7.2 cannot by built on windows.  However, we must add `iris==1.7.2_DEV_9e1fd13` to our requirements file on Linux to avoid installing the latest version until this bug is fixed:

https://github.com/SciTools/iris/pull/1502

and live with the bug on windows...

### DIFF
--- a/iris/bld.bat
+++ b/iris/bld.bat
@@ -1,0 +1,5 @@
+set SITECFG=lib/iris/etc/site.cfg
+echo [System] > %SITECFG%
+echo udunits2_path = %SCRIPTS%\udunits2.dll >> %SITECFG%
+
+%PYTHON% -sE setup.py install --single-version-externally-managed --record record.txt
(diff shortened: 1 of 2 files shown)


rsignell-usgs:
Do we have to bump the build number to try this again?

ocefpaf:
Not in this case.  There is no build zero for iris 1.7.4 on linux, so it will get build once the PR ismerged (I hope!).  And, on windows, this package is already on ioos/binstar due to my AppVeyor workaroun, that means: bumping the build number for windows would duplicate the binary on binstar.

rsignell-usgs:
So should I merge this?

ocefpaf:
Yes.  At least at very least, by merging this PR, we will make the binstar channel match with the Github repository.

PR #84, opened 2015-02-20
Re-add shapely.

ocefpaf:
Shapely is not available on windows in the default channel.  My conda config had the scitools channel added when I proposed the deprecation and made me think we had wind32/64 packages.  This PR re-adds the package.

@rsignell-usgs sorry for the mess.


pelson:
Question: What scope do you have for this repository? For example, if the scitools channel already provides a distribution, is that something that you want to just make use of, or would you rather maintain your own recipe?

ocefpaf:
That is a good question!  I am pinging @rsignell-usgs to help answer it.

Here is what I can say:  First I wanted to add scitools/conda-recipes-scitools as a submodule here.  That turned out to be a bad because we had to diverge our iris version.  We are currently using iris 1.7.2 from a branch that fixes https://github.com/SciTools/iris/pull/1502.

Then I thought about adding scitools channel to our tutorials.  But having everything in one channel is easier for most users and avoid mixing matching channels.  (Look at myself.  I just made a mistake here because of adding/removing channels with care.  Conda should really have a feature to isolate channels for each env.)

rsignell-usgs:
Yes, the main problem is that since we can't use the SciTools Iris in it's current state, we have to build our own.  And if we build our own, it seems that it's safer just to build the dependencies.    Since travis and appveyor are doing the work, who cares, right?

pelson:
> Since travis and appveyor are doing the work, who cares, right?

Right - except there is an overhead to maintaining the recipes, and personally I'd sooner share the burden than have us duplicate the work. Perhaps its something we can discuss at SciPy this year - maybe we would want to share built distros over multiple channels or even just combine the repos altogether. Either way - no rush, just something for us to think about.

ocefpaf:
@pelson I am worried about that too.  (Specially because Rich and I are the only ones maintaining this for now :stuck_out_tongue_winking_eye:)

Combining repos might be a burden on your side.  Our repository is not as professional as [conda-recipes-scitools](https://github.com/SciTools/conda-recipes-scitools).  Sharing built distros might be a good idea.  However, conda's current model to copy another channel binary is a trick that leaves no tracks for reproducibility.  The recipes here are our bread crumbs.  If there is a way to create a recipe that copies the binary...

BTW I was planning on re-visiting the submodules idea once https://github.com/SciTools/iris/pull/1502 gets sorted out.

PS: I am not going to SciPy :frowning:, but I will be at Pycon.

PR #85, opened 2015-02-20
Fix win32 build.

### LINKED ISSUE #81: Build psycopyg2 on win32
Option on the table:

- Build from an specific branch (ssh://hg@bitbucket.org/topper/psycopg2-win32-py27)
- Copy an existing binary
- Bite the bullet and try to build the latest version from source

ocefpaf:
This PR adds an IF/ELSE clause in the `bld.bat` to install the proper `psycopg2` binary for each ARCH.  Because the binary available for win32 is 2.5.2 I downgraded the package to that version.

(`pycsw` is the only package that uses `psycopg2`.)

This PR closes #81.

### DIFF
--- a/psycopg2/bld.bat
+++ b/psycopg2/bld.bat
@@ -1,1 +1,5 @@
-pip install psycopg2
+IF %ARCH% == 32 (
+   pip install git+https://github.com/nwcell/psycopg2-windows.git@win32-py27#egg=psycopg2
+) ELSE (
+   pip install psycopg2==2.5.2
+)
--- a/psycopg2/meta.yaml
+++ b/psycopg2/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: psycopg2
-  version: "2.5.4"
+  version: "2.5.2"
 
 requirements:
   build:


rsignell-usgs:
From @tomkralidis: "pycsw only needs psycopg2 if the backend is PostgreSQL"

ocefpaf:
Good to know.  Maybe we could suggest them to make it a soft dependency then.

tomkralidis:
psycopg2 is a soft dependency for pycsw and is not included in pycsw's setup.py's use of https://github.com/geopython/pycsw/blob/master/requirements.txt

ocefpaf:
Ops.  So we have to remove it from out conda recipe.  What do you think @rsignell-usgs?

@tomkralidis sorry for not looking into it before!

tomkralidis:
No problem glad to help! Hopefully this makes things just a bit lighter here.

rsignell-usgs:
I forgot how it got into our meta.yaml in the first place.  Maybe from pypi?   Yes, let's remove it.

ocefpaf:
And remove `psycopg2` as well?  I do not use the PostgreSQL backend, so I not the right guy to say so.

PR #88, opened 2015-02-21
remove psycopg2 dependency

tomkralidis:
psycopg2 is not required by pycsw unless pycsw is using a PostgreSQL backend, which doesn't appear to be the case here.

Inspired by https://github.com/ioos/conda-recipes/pull/85 dicsussion.


ocefpaf:
Awesome!  Thanks @tomkralidis!!

@rsignell-usgs Do you want to remove the `psycopg2` binaries from binstar (the command is `binstar remove ioos/psycopg2`) or should we leave them there as a fallback for people trying to use the  PostgreSQL backend?

@tomkralidis We are using [Obvious-CI](https://github.com/pelson/Obvious-CI) to build the packages and that requires a bump in the build number and/or a package version update to get a new pycsw build.  Right no build will happen and the binary hosted on binstar still uses `psycopg2`.  (Sorry for that!  We are still writing some docs for this repository!)

PR #89, opened 2015-02-21
Bump build number.

ocefpaf:
This complements PR #88.

### DIFF
--- a/pycsw/meta.yaml
+++ b/pycsw/meta.yaml
@@ -6,7 +6,7 @@ source:
   git_url: https://github.com/geopython/pycsw.git
 
 build:
-  number: 0
+  number: 1
   preserve_egg_dir: True
 
 requirements:


ocefpaf:
Just to illustrate this.  (We can use this thread as an example later, when writing some docs.)

PR #88 did not bump the build number:
https://travis-ci.org/ioos/conda-recipes/builds/51626433#L323

PR #89 incremented the build number:
https://travis-ci.org/ioos/conda-recipes/builds/51634104#L323

rsignell-usgs:
Is there any limit on the build number?  :smiley_cat:

ocefpaf:
In theory no.  But the best practice is to reset the **build number** when updating the **version number** of a package.

Higher package version takes precedence over lower package version no matter the build number of the lower version number.  Higher build number takes precedence over lower build number only for the same package version.

rsignell-usgs:
Okay, thanks for the explanation!

PR #93, opened 2015-02-24
removing prettyplotlib, as not needed

rsignell-usgs:


00_env_requirements/ioos/windows64_req.txt
```diff
@@ -6,6 +6,5 @@ folium
 rdflib
 geojson
 requests
-prettyplotlib
```
ocefpaf: `prettyplotlib` was there because some notebooks on system-test still use it.  But I think it is best to let them break and fix it as people try to run them.

PR #96, opened 2015-02-25
we only need one requirements file for linux

rsignell-usgs:


ocefpaf:
@rsignell-usgs maybe we should consider renaming `linux_req.txt` to `req.txt`, since it will be common to MacOS and only windows will be different.

rsignell-usgs:
okay, I'll do that

ocefpaf:
@rsignell-usgs I am going to merge this as is and we can do that once we have OSX packages on binstar and tested.

PR #98, opened 2015-03-03
re-wrote libspatialindex to download source tarball, rather than pull from gitHub

ChrisBarker-NOAA:
This is so that autoconf will have already been run -- so you can build with a simple:

./configure && make && make install

Note that OS-X doesn't have autotools installed by default, so this makes it a lot easier to build on OS-X

And it seems the "official" source tarball is probably the way to go anyway.

### DIFF
--- a/libspatialindex/build.sh
+++ b/libspatialindex/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #cmake -DCMAKE_INSTALL_PREFIX=$PREFIX .
-./autogen.sh 
+#./autogen.sh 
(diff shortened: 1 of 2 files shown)


libspatialindex/build.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 #cmake -DCMAKE_INSTALL_PREFIX=$PREFIX .
```
rsignell-usgs: If there is an option to build packages with CMAKE, shouldn't we be using it?

PR #101, opened 2015-03-03
Install gfortran on Linux only.

ocefpaf:
- Installs gfortran (via apt-get) only for Linux.
- Removed sudo calls (not need on travis-ci)


ocefpaf:
@rsignell-usgs I just rebased this PR and it is safe to merge.  It will avoid further headaches down the road when we get a MacOS travis slot.

PR #102, opened 2015-03-03
Updated `pytest-cov` and `cov-core`.

ocefpaf:
Updated `pytest-cov` and its dependency `cov-core`.


ocefpaf:
@rsignell-usgs Another safe merge, just updating the versions.

I would like to see these two on conda default... Meanwhile lets keep them here.

PS: The AppVeyor failure is fixed by #110

PR #105, opened 2015-03-04
Added flags to the setup.py install command so that setuptools wouldn't ...

ChrisBarker-NOAA:
...check for dependencies.

It was having trouble finding petulant-bear, even though it was there.

NOTE: setuptools is often a pain in this way -- it expects eggs, etc, and conda overrides that. I"m guessing tha the hyphen in petulant-bear is messing things up...

I got this solution from:

https://groups.google.com/a/continuum.io/forum/#!topic/conda/3Jw2cnsL_v8

### DIFF
--- a/wicken/build.sh
+++ b/wicken/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-$PYTHON setup.py install
+$PYTHON setup.py install --single-version-externally-managed --record=/tmp/record.txt
 
 # Add more build steps here, if they are necessary.
 


ocefpaf:
The AppVeyor failure is unrelated to your changes and PR #103 fixes it.

 @ChrisBarker-NOAA can you review and merge #103?  @rsignell-usgs us out and I try not to do self merges.

Also, can you bump the build number on `wicken`, so it gets a new build for all systems with the new option.

PR #107, opened 2015-03-05
Updated udunits build to use official tarball from unidata.

ChrisBarker-NOAA:
Updated udunits build to use official tarball from unidata, rather than gitHub repo.

This gets you a version with a configure script ready to go.

Also updated to newer version: 2.2.18 -- 'cause why not?

### DIFF
--- a/udunits2/build.sh
+++ b/udunits2/build.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 
-if [ ! -f configure ];
-then
-   # Make the configure file. Need autoreconf, libtool, libexpat-dev for this.
-   autoreconf -i --force
-fi
+## should have a configure file if you got the offical source tarball
+# if [ ! -f configure ];
+# then
+#    # Make the configure file. Need autoreconf, libtool, libexpat-dev for this.
+#    autoreconf -i --force
+# fi
 
 ./configure --prefix=$PREFIX
 make
--- a/udunits2/build.sh
+++ b/udunits2/build.sh
@@ -1,11 +1,13 @@
 #!/bin/bash
 
-## should have a configure file if you got the offical source tarball
-# if [ ! -f configure ];
-# then
-#    # Make the configure file. Need autoreconf, libtool, libexpat-dev for this.
-#    autoreconf -i --force
-# fi
+# should have a configure file if you got the offical source tarball
+# but if it got pulled from the git repo, the configure script needs to be 
+# rebuilt
+if [ ! -f configure ];
+then
+   # Make the configure file. Need autoreconf, libtool, libexpat-dev for this.
+   autoreconf -i --force
+fi
 
 ./configure --prefix=$PREFIX
 make
--- a/udunits2/meta.yaml
+++ b/udunits2/meta.yaml
@@ -1,11 +1,10 @@
 package:
     name: udunits2
-    version: "2.2.17"
+    version: "2.2.18"
 
 source:
-    fn: v2.2.17.tar.gz
-    url: https://github.com/Unidata/UDUNITS-2/archive/v2.2.17.tar.gz
-    sha1: 3dfb7b1f18e21ac5366127778bbfd50ab5b7cf0a
+    fn: udunits-2.2.18.tar.gz
+    url: ftp://ftp.unidata.ucar.edu/pub/udunits/udunits-2.2.18.tar.gz
 
 requirements:
   build:


ocefpaf:
@pelson  before merging this here I would like to ask you if there was any special reason for using the GitHub tarball instead of the one from the ftp?  Or was it just the fact unidata selective removed some of old versions from the ftp?

ChrisBarker-NOAA:
By the way -- I posted an issue on gitHub about the configure script before I found the tarball -- they are very clear that they will NOT put a configure script in the repo.

Though I see that we were pulling from "archive" -- not sure what that means, but it might be reasonable to put a configure script in an archive. But the official tarballs seem to make the most sense to me.

NOTE: what's with the patchbinary.py stuff? That sure seems like something that conda should be doing for us, or not needed at all...

-CHB

ocefpaf:
AppVeyor is not happy with change...  I will investigate this tomorrow morning.

pelson:
> before merging this here I would like to ask you if there was any special reason for using the GitHub tarball instead of the one from the ftp

It kept getting deleted. See https://github.com/SciTools/conda-recipes-scitools/issues/17 and the straw that broke the camel's back https://github.com/SciTools/conda-recipes-scitools/pull/16.

ocefpaf:
Thanks @pelson!

@ChrisBarker-NOAA I am reluctant to merge this until we fix the windows build.  And since you are using the TOKEN locally the package for OSX is already in the binstar channel, so no harm for Mac user if we wait.

ChrisBarker-NOAA:
Agreed -- it shouldn't be merged until it passes everywhere -- I wonder what's up with the Windows build?

But if unidata keeps pulling older source tarballs from their ftp site -- then we're kind of stuck. 

I did post a whine on the issue on gitHub, but who knows -- they apparently have _some_ reason to only put the latest minor builds up -- though I have no idea what that is.

The thing is, I'd really rather not have autotools be required to build the conda package if we can avoid it.

ocefpaf:
> The thing is, I'd really rather not have autotools be required to build the conda package if we can avoid it.

Makes sense!

ChrisBarker-NOAA:
HMM -- can you give two locations in the meta.yaml -- one to look first, and the second to use as a fallback? That could help.

ocefpaf:
Not sure if it is possible, but if it is I do not encourage that practice.  I like to have one canonical source for tarballs to avoid confusion.  And if the main source of a file breaks we should know why.

ChrisBarker-NOAA:
So unidata says they only want to keep the latest third-point version on their site so that folks won't get a version with bugs that have been fixed. I think it's pretty standard practice to keep old builds around, but it may be good that it will break when a bug-fix release comes out -- then we know we need to go update the version.

not sure if conda has a way to spell :file_name.X.Y.largest in this context.

But in any case, my vote is to stick with the ftp site for the source, at least if we can get it to build with Windows.

ocefpaf:
> So unidata says they only want to keep the latest third-point version on their site...

If by site they meant website I'd agree.  But it is just an ftp!  old versions, even buggy, ones are needed for the sake of [reproducibility](http://xkcd.com/1172/).

> But in any case, my vote is to stick with the ftp site for the source, at least if we can get it to build with Windows.

Agreed.  I will working on the windows package this weekend.

ChrisBarker-NOAA:
On Thu, Mar 5, 2015 at 1:55 PM, Filipe <notifications@github.com> wrote:

> So unidata says they only want to keep the latest third-point version on
> their site...
>
> If by site they meant website I'd agree. But it is just an ftp! old
> versions, even buggy, ones are needed for the sake of reproducibility
> <http://xkcd.com/1172/>.
>

I agree, and made the case in the gitHub issue....but I'm not optimistic
that they are going to change their mind. They may add a "latest" link,
though.

Ironically, the website is actually out of date with the version number....

-Chris





>  But in any case, my vote is to stick with the ftp site for the source,
> at least if we can get it to build with Windows.
>
> Agreed. I will working on the windows package this weekend.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ioos/conda-recipes/pull/107#issuecomment-77460557>.
>



-- 

Christopher Barker, Ph.D.
Oceanographer

Emergency Response Division
NOAA/NOS/OR&R            (206) 526-6959   voice
7600 Sand Point Way NE   (206) 526-6329   fax
Seattle, WA  98115       (206) 526-6317   main reception

Chris.Barker@noaa.gov

pelson:
> I agree, and made the case in the gitHub issue....but I'm not optimistic that they are going to change their mind. They may add a "latest" link, though.

For me, the canonical source of the source is the source. That is now (thankfully) on github, so I strongly feel that whatever is necessary to be done to build from source is what needs to be done to build from source.

I'm in a repetitive mood today. I'm in a repetitive mood today. :smile:

ocefpaf:
@pelson you say that because you have a Travis-ci with OSX and you don't have to configure anything!  Wait until we get ours!!! :stuck_out_tongue_closed_eyes:

ocefpaf:
Closing this see #143.

PR #108, opened 2015-03-09
New token

ocefpaf:
The Windows binaries are not uploaded to binstar when merging, only when build from @ocefpaf account.

This PR replaces the APPVeyor TOKEN in an attempt to fix it.  The test package is `oceans`, it should get build in the PR, but not uploaded!  The upload should happen only when merging.


ocefpaf:
@rsignell-usgs OK, the PR build was successful: `oceans` was built but not uploaded as expected.  Now can you merge.  The next step is to see if the merge will activate the main repository build and that one should upload!

ocefpaf:
@rsignell-usgs OK, that worked!

PR #122, opened 2015-03-19
adding dependencies to utilities

rsignell-usgs:


ocefpaf:
@rsignell-usgs I don't think we should move it to the deprecation directory.  We can get health builds for Linux (other than RHEL) and OSX.  I think that it should be the case to just: "do not build not RHEL." (Like we do for Windows.)

rsignell-usgs:
@ocefpaf, okay, how do I do that?

ocefpaf:
Sadly there is no way to do it here.  But since it is a one time build, just delete the directory locally and fire obvious-ci again.

rsignell-usgs:
@ocefpaf, yeah, switching to the @ChrisBarker-NOAA version of `udunits2/meta.yaml` broke the windows builds on appveyor.    But it does seem like `utilities/meta.yaml` was missing run dependencies on `requests` and `owslib`.   How do you want to handle?

utilities/meta.yaml
```diff
@@ -22,6 +22,8 @@ requirements:
     - python
     - iris
     - lxml
+    - owslib
+    - requests
```
ocefpaf: `iris` installs `cartopy` that install` owslib`
`cartopy` installs `pyepsg` that install `requests`.

So I cannot see how the dependencies are missing.  Unless `iris` and cartopy are coming from a channel other than ours.

ocefpaf:
@rsignell-usgs Did you change the name and purpose if this PR?  I thought it was about deprecating rasterio or not.  I am lost in the commits here.

ocefpaf:
All that said, it is not a problem to over specify dependencies like `owslib` and `requests` in `utilities`.  Maybe it is better to do it since both `owslib` and `pyepg` (that installs `requests`) are optional dependencies to cartopy and people using custom installations might be missing them.

Can you send those changes in another PR?  Or do you want me to do it?  I do not recommend merging this PR because it brings @ChrisBarker-NOAA changes to `udunits2` that are not ready to be merged yet (the windows failure).

rsignell-usgs:
@ocefpaf , okay, I have to get better with these PRs.   Yes, I changed the name because first I deprecated rasterio, and then I put it back on your recommendation, but also added those requirements to utlities, and not only that, changed to udunits from @ChrisBarker-NOAA , which broke appveyor.  So it really should have been two PRs, but apparently both should be rejected.

utilities/meta.yaml
pelson: You will probably want to increment the build number?

ocefpaf: @rsignell-usgs yep, when the version is not changing we have to remember to increment it.  (But `utilities` version will change soon :smile: https://github.com/pyoceans/utilities/pull/34)

@pelson Just an idea.  I build a lot of stuff using OpenSUSE build service and the system compare the spec file changes, using a version control system, and increment the build number automatically.  Maybe that could be added to obvious-ci?  Or maybe it is a lot of work for such a simple thing...

ocefpaf:
@rsignell-usgs don't worry about the PR.  We don't have to be so strict here :wink:

utilities/meta.yaml
pelson: It might be a bit of work yes, but could be worthwhile. What interface did you have in mind?

ocefpaf: Not sure.  In OpenSUSE the solution was a tool called (osc) builds and does the version control of the packages.  In our case `obvious-ci` abstracts `conda-build` and `binstar`, and the version control is "optional" (~most of the times~ GitHub).

I can only thing of some hackish solutions where obvious would compare the `meta.yaml` file using some git magic and, if different, increments the build number to triggers a new build.

PR #127, opened 2015-03-24
Added pymatbridge.

ocefpaf:
@rsignell-usgs I have only one notebook that calls matlab :grin:, but I believe this might be important for other cases.

### DIFF
--- a/pymatbridge/bld.bat
+++ b/pymatbridge/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
--- a/pymatbridge/build.sh
+++ b/pymatbridge/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install
+
+# Add more build steps here, if they are necessary.
+
+# See
+# http://docs.continuum.io/conda/build.html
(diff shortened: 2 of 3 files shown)


rsignell-usgs:
@ocefpaf , I agree.    We have a lot of matlab users, and this should be useful.

ocefpaf:
Latest master fixes pymatbridge for win64, but it is not available for win32.  I will modify the recipe to pull from master.  Do not merge it yet.

rsignell-usgs:
Just as a reminder, the reason we still want the 32-bit windows binaries is to interface with python-based apps like ESRI 32-bit ArcMap.

ocefpaf:
Don't worry, I will not remove our win32 build on pymatbridge account :smile:

rsignell-usgs:
@ocefpaf , strange.  We expected the 32bit to fail, right?  But the 64-bit is failing too, with an error saying that pymatbridge doesn't work on *32 bit*:
https://ci.appveyor.com/project/comtbot/conda-recipes/build/1.0.114/job/0hu61er5fqcggkox#L984

ocefpaf:
(I think you missed my message above :wink:)

> Latest master fixes pymatbridge for win64

This is a bug in `pymatbridge` that is fixed latest master, I just need to point our recipe to it.  (In fact, doing it now.)

rsignell-usgs:
@ocefpaf ,  I saw your message.   I just thought that was already modified.   I should have looked at the meta.yaml!

ocefpaf:
@rsignell-usgs is should be good to go now.  Sorry it took so long for a simple thing.  I am :sleeping:  today.

PR #169, opened 2015-04-03
Fixed geolinks

ocefpaf:
@rsignell-usgs this is not an updated!  The metadata seemed to be from another package (version, URL, etc).  Can you review if this is the right package?

### DIFF
--- a/geolinks/bld.bat
+++ b/geolinks/bld.bat
@@ -1,8 +1,2 @@
 "%PYTHON%" setup.py install
 if errorlevel 1 exit 1
-
-:: Add more build steps here, if they are necessary.
-
-:: See
-:: http://docs.continuum.io/conda/build.html
-:: for a list of environment variables that are set during the build process.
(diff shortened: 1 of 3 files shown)


rsignell-usgs:
Yes, this is the right package, but requesting a release here:
https://github.com/geopython/geolinks/issues/1

ocefpaf:
> but requesting a release

Lets put this PR on hold then.

rsignell-usgs:
Actually, I guess since @tomkralidis has a versioned pypi package, we don't need the tagged github release.

ocefpaf:
I am OK with it. :+1:

tomkralidis:
https://github.com/geopython/geolinks/issues/1#issuecomment-89465686



FYI we had 0.0.1 in PyPI, and there have some (minor) changes since then, but no tag.

For good measure I have cut a 0.1.0 release tag and updated PyPI as well.

PR #175, opened 2015-04-03
Using tag instead of master

ocefpaf:
@phobson is this tag OK with the development branch?  Should we use it?


phobson:
Can you give me until after the weekend to add a few more tests and some coverage reports to the repo? Then I'll feel more comfortable to do a proper release like the dependencies.

ocefpaf:
Sure.  No problem @phobson.  We are not under a deadline here.  Just updating our conda packages :wink: 

Thanks!

PR #180, opened 2015-04-06
Updated pyGDP

ocefpaf:
- Added a requirements specifically for running tests.
- Bug fix on _generateFeatureRequest function.


rsignell-usgs:
@dblodgett-usgs, just FYI, this means the latest pygdp is available for anaconda users on all platforms at:
https://binstar.org/ioos/pygdp

PR #187, opened 2015-04-14
bump cartopy to 0.12.0

rsignell-usgs:


rsignell-usgs:
@ocefpaf , cartopy 0.12.0 wants shapely>1.55 and shapely currently is at
https://github.com/Toblerity/Shapely/releases/tag/1.5.7
Any reason not to use bump that?

pelson:
Potentially for the OSX issue... https://github.com/Toblerity/Shapely/issues/177
I'm hoping to do a little bit of digging to see if that issue has been resolved...

ocefpaf:
> Potentially for the OSX issue... Toblerity/Shapely#177

Yes, that's is the reason we pinned the version (and ditched the conda default).  We where basically following Scitools recipes.

I believe the reason why cartopy 0.12 needs a newer shapely is a bug in shapely and not new features.  This bug could be [circumvented](https://github.com/SciTools/cartopy/issues/597) in cartopy itself, and we could still use the current version.  Although I agree that fixing shapely is best in the long term.

> I'm hoping to do a little bit of digging to see if that issue has been resolved...

Thanks @pelson.

BTW: I am to writing a similar patch for `rtree` as well.

ocefpaf:
@rsignell-usgs I re-started AppVeyor and cartopy  is OK with our new Shapely.  Merging this...

rsignell-usgs:
Does this mean I should update the RHEL6 also?

ocefpaf:
Yes.  And I am updating OSX.

(We need a better solution :unamused:)

Issue #185, opened 2015-04-12
mingwpy


ocefpaf:
@rsignell-usgs and @pelson

This project https://bitbucket.org/carlkl/mingw-w64-for-python/downloads is moving to github https://github.com/mingwpy.  (And soon it will be getting some "wheels" and conda)

We should check it out to get gcc, gfortran and etc for Windows.

See https://us.pycon.org/2015/schedule/presentation/377/

pelson:
Cool! Thanks for sharing @ocefpaf.

rsignell-usgs:
@ocefpaf , the abstract says "show how to embed third party dynamic libraries (.dylib or .dll) in a wheel package on OSX and Windows to make it independent of non-Python dependencies,"   

Does that mean that Wheels can have all the functionality of Conda?

ocefpaf:
@rsignell-usgs that means conda has all the functionality of wheels :wink: 
However, wheels binaries are not relocatable like conda binaries.

ocefpaf:
BTW, just out of curiosity, Oliver Grisel showed me an all Linux solution for windows packages using wine+mingwpy instead of AppVeyor!

pelson:
> Oliver Grisel showed me an all Linux solution for windows packages using wine+mingwpy instead of AppVeyor!

Goodness. I can feel the maintenance headache already :smile: 

> Does that mean that Wheels can have all the functionality of Conda?

I'm still a little bothered that we will end up with big wheels that ship all of the binaries that a package depends on, only to duplicate the binaries in other wheels (e.g. shapely and cartopy both depend on GEOS). That said, there are some benefits from the wheel approach - most notably for me is the ability to pip install into any Python installation (even conda), where conda distributions can only be installed (comfortably) into conda environments.

ocefpaf:
> Goodness. I can feel the maintenance headache already

You have to admit that was a creative solution :wink:,  but I believe we are OK with AppVeyor.  Their reason for looking into this is speed.  AppVeyor is slow for their use case.  We are just shipping while they are build+testing every PR.

> duplicate the binaries in other wheels

Yes, PyPI+pip should think ~~out~~inside the box and do what other package managers already do.  Or just reduce its use to pure python packages.

ocefpaf:
PR #393 closes this.  Now I will tackle one-by-one of the Python modules that need gfortran/gcc.  @rsignell-usgs `octant` is already in!

PR #189, opened 2015-04-15
Sci-wms repo and recipe update.

### LINKED ISSUE #173: Which repository should we use for sci-wms?
https://github.com/kwilcox/sci-wms.git or https://github.com/sci-wms/sci-wms

ocefpaf:
This PR closes #173.

@rsignell-usgs and @kwilcox I would like to:
- [ ] Write a script to "copy" `sci-wms` to a working directory and drop the name  `sci-wms-deps`
- [ ] Run a test on that working directory (like: `python manage.py test`).

What do you think?  Worth pursuing? This script would be something like `sci-wms [new|deploy|help|serve]` format.

### DIFF
--- a/djangorestframework/bld.bat
+++ b/djangorestframework/bld.bat
@@ -1,0 +1,2 @@
+"%PYTHON%" setup.py install --single-version-externally-managed --record record.txt
+if errorlevel 1 exit 1
--- a/djangorestframework/build.sh
+++ b/djangorestframework/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install --single-version-externally-managed --record record.txt
--- a/djangorestframework/meta.yaml
+++ b/djangorestframework/meta.yaml
@@ -1,0 +1,32 @@
+package:
+    name: djangorestframework
+    version: "3.1.1"
+
+source:
+    fn: djangorestframework-3.1.1.tar.gz
+    url: https://pypi.python.org/packages/source/d/djangorestframework/djangorestframework-3.1.1.tar.gz
+    md5: d0f6aa82d49e63faebaadfc089e5e80b
+
+build:
+    number: 0
+
+requirements:
+    build:
+        - python
+        - setuptools
+    run:
+        - python
+
+test:
+    imports:
+        - rest_framework
+        - rest_framework.authtoken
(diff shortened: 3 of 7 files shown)


rsignell-usgs:
Andrew (@ayan-usgs) or Chris (@cbcunc), can you respond to Filipe (@ocefpaf)?    
@kwilcox is going to be gone for a few weeks (baby #3) and I'm not really qualified to answer.

ocefpaf:
The Windows failure is because `setproctitle` needs a C compiler.  I will disable it on windows for now until we have #185 working.

rsignell-usgs:
@ocefpaf, having sci-wms on windows is not a high-priority (although I think @ayan-usgs is running sci-wms on windows)

ocefpaf:
This will not disable `sci-wms` on windows.  It is only the `setproctitle` dependency that will not be satisfied.   (I am not a user to say what will be the side effects of that.)

ayan-usgs:
@rsignell-usgs  and @ocefpaf,

I was unable to install setproctitle in a Python 2.7 virtual environment using Visual Studio 2008. However, in reference to #185, I found that the install does work  when using mingw32 as the compiler. 

I've had good success with mingw32 for awhile now, though with the caveat that some packages from pypi can be installed using VS 2008 or mingw32, while others can't. I still find myself periodically going to http://www.lfd.uci.edu/~gohlke/pythonlibs/ to find a wheel or an executable for those times where it's not worth fighting with Windows.

ocefpaf:
> I found that the install does work when using mingw32 as the compiler.

Good to know!  I am working to get this going in our AppVeyor setup.

Another question: is `setproctitle` a mandatory or optional dependency for sci-wms?  (The current docs seems to be outdated and it is hard to know the details.)

rsignell-usgs:
@ocefpaf good to go on this, or wait?

ocefpaf:
If `setproctitle` is optional we are good to go.  But the docs are unclear about that...

ayan-usgs:
@ocefpaf,

'setproctitle' is not actually needed to run sci-wms.

ocefpaf:
OK, I got it from here:

https://github.com/sci-wms/sci-wms/blob/master/requirements-prod.txt#L2

@rsignell-usgs this PR is good to go then.

PR #190, opened 2015-04-15
Updating Shapely and patch.

ocefpaf:
This should fix #187 (See https://github.com/SciTools/conda-recipes-scitools/pull/68/)


ocefpaf:
Note to self:  We need to update `geos_c.win.patch` as well.  I will look into this tomorrow morning.

ocefpaf:
@rsignell-usgs This is good to go.  After this we can rebase #187.

PR #194, opened 2015-04-22
Adding mo_pack

ocefpaf:


pelson:
I can't wait until we have a better solution for this sharing :wink:
I've got something lined up that I want to put some finishing touches to, and would like to talk it through with you. Problem is, its been too :sunny: :sunglasses: here to work on it in the evenings :beers: ... forecast is looking good for development this weekend :wink:

ocefpaf:
> I can't wait until we have a better solution for this sharing

Me too.  I feel like a package :monkey: copying-and-pasting recipes!

> I've got something lined up that I want to put some finishing touches to, and would like to talk it through with you.

I'd love that!

Regarding the sun... Well, enjoy it as much as you can.  But I live in a place where we run **from the :sunny:  and not to it**.  The year round forecast here is min: 27 degC max: 29 degC with chances of :umbrella:

PR #200, opened 2015-04-27
Removed unnecessary header file.

ocefpaf:
Udunits 2.2.19 does not need `unistd.h`


ocefpaf:
@rsignell-usgs this is a harmless one.  So I am self-merging.

rsignell-usgs:
Okay!

PR #203, opened 2015-04-28
Re-build cartopy with proj4-4.9.1

ocefpaf:
@rsignell-usgs If a user install the latest proj4 with the current cartopy s/he may get an error like

```python
ImportError: libproj.so.0: cannot open shared object file: No such file or directory
``` 

Re-building it against latest `proj4` should solve the issue.


ocefpaf:
Self-merging this one...

PR #214, opened 2015-05-04
Bump the build number for clean Linux/OSX binaries

ocefpaf:
@rsignell-usgs Do not merge this one yet.  Most of this PR are the build number "bumps" I had to do to get a complete Linux re-build.

I was thinking that, since the channel has  OSX binaries from 3 different machines (Chris Barker, Rusty, and Travis), maybe it is good to test a complete re-build for OSX too.

AppVeyor will  crash due to time limit, but we can ignore that.

### DIFF
--- a/affine/meta.yaml
+++ b/affine/meta.yaml
@@ -1,27 +1,27 @@
 package:
-  name: affine
-  version: "1.1.0"
+    name: affine
+    version: "1.1.0"
 
 source:
-  fn: affine-1.1.0.tar.gz
-  url: https://pypi.python.org/packages/source/a/affine/affine-1.1.0.tar.gz
-  md5: a3d020a8f0ebb762294caf0cc5d28d18
+    fn: affine-1.1.0.tar.gz
+    url: https://pypi.python.org/packages/source/a/affine/affine-1.1.0.tar.gz
+    md5: a3d020a8f0ebb762294caf0cc5d28d18
 
 build:
-  number: 0
+    number: 1
 
 requirements:
-  build:
-    - python
-    - setuptools
-  run:
-    - python
+    build:
+        - python
+        - setuptools
+    run:
+        - python
 
 test:
-  imports:
-    - affine
+    imports:
+        - affine
 
 about:
-  home: https://github.com/sgillies/affine
-  license:  BSD License
-  summary: 'Matrices describing affine transformation of the plane'
+    home: https://github.com/sgillies/affine
+    license:  BSD License
+    summary: 'Matrices describing affine transformation of the plane'
--- a/airsea/meta.yaml
+++ b/airsea/meta.yaml
@@ -1,28 +1,28 @@
 package:
-  name: airsea
-  version: !!str 0.0.1_DEV
+    name: airsea
+    version: "0.0.1.dev"
 
 source:
     git_url: https://github.com/ocefpaf/python-airsea.git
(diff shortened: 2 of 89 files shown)


rsignell-usgs:
Okay, should I wait for clean linux binaries also before testing them on the comt platform?

ocefpaf:
Nope, the Linux binaries are already there.  Go ahead and let me know if they work for you.  (I sent you an e-mail about it.)

ocefpaf:
I am not sure why conda is pulling up geos 3.3.3 instead of ours (3.4.2), but that is breaking cartopy.

rsignell-usgs:
The linux version on binstar is working fine though, right?

ocefpaf:
The right answer is yes an no.  The Linux builds are also pulling geos 3.3.3, but the crash only happens on  OSX.  Linux cartopy works just fine.  (See issue https://github.com/SciTools/cartopy/issues/612)

ocefpaf:
@rsignell-usgs My guess is that geos 3.3.3 is from basemap.  We have a few packages that are pulling basemap (octant and sci-wms).

One solution would be to stop using basemap from the default channel and provide one with geos 3.4.2.  However, conda should be smart enough to update geos.  (Unless their basemap package is pinned the geos version.)

rsignell-usgs:
Interesting. Ideally we would get rid of basemap in favor of cartopy. I know that basemap is in the process of being dropped my sci-wms and I'm not sure how much octant needs it.

ocefpaf:
@rsignell-usgs This is failing due to time-out, but after #216 and #215 are merged this PR should  be OK.  After the merge I can keep "re-starting" travis and AppVeyor to get all the packages built and uploaded.

shoyer:
I think you need to rebuild basemap to get it to work with geos 3.4.2 (needed for cartopy). The good news is that the recipe is not very complicated: https://github.com/TheClimateCorporation/conda-recipes/blob/master/basemap/meta.yaml

ocefpaf:
Thanks @shoyer I actually already have that in there: https://github.com/ocefpaf/conda-recipes/blob/Linux_builds/basemap/meta.yaml#L16

This PR is massive and ill formed.  I needed it to re-build all the Linux64 binaries using a docker  image.  the right thing to do would be to wipe the binaries from the channel and start over, but I do not want to loose the download statistics so I bumped all the build numbers.

The current failure is due to travis and AppVeyor timing-out.  Things are consistent with geos-3.4.2 now.  and this PR is safe to be merged.

BTW: The  main issue was not basmap, it was the default anaconda shapely. (See #216.)

rsignell-usgs:
@ocefpaf , this one can't be automatically merged.

ocefpaf:
@rsignell-usgs rebased and ready to go.  We can wait for travis, but it will probably time-out again.

rsignell-usgs:
Merging because errors are just timeouts

rsignell-usgs:
Interesting. Pull requests with failing status can't be merged from a phone!

ocefpaf:
I will keep "re-starting" Travis and AppVeyor once this PR lands on `master`.  Every re-start it will take less time because the packages **are** uploaded after the merge.

PS: In the future we should avoid doing something like this again.  It is preferable to bump every package in a separate PR.

ocefpaf:
Let me try then...  Done!

PR #215, opened 2015-05-05
Updated oceans

ocefpaf:
Updated to latest release

### DIFF
--- a/oceans/bld.bat
+++ b/oceans/bld.bat
@@ -1,8 +1,1 @@
-"%PYTHON%" setup.py install
-if errorlevel 1 exit 1
-
-:: Add more build steps here, if they are necessary.
-
-:: See
-:: http://docs.continuum.io/conda/build.html
-:: for a list of environment variables that are set during the build process.
+"%PYTHON%" setup.py install install --single-version-externally-managed  --record record.txt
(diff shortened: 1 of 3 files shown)


rsignell-usgs:
So oceans was the culprit for geos 3.3?
Do you need to restart appveyor?

ocefpaf:
Nope, but oceans calls cartopy and it was failing when cartopy pulled geos 3.3.3

I am puzzled by how conda deals dependencies.  I cannot see why geos 3.3.3 gets precedence over the geos 3.4.2 in the ioos channel.

rsignell-usgs:
@ocefpaf , how about asking the conda google group, using this exact example to illustrate the problem?

ocefpaf:
> Do you need to restart appveyor?

Nope.  Win32 worked just fine and Win64 is in the Queue.

>  How about asking the conda google group, using this exact example to illustrate the problem?

I have a draft e-mail here.  I am just waiting for these tests to finish.

ocefpaf:
@rsignell-usgs I finally found the source of the issue!  Conda updated its default shapely to `1.5.8` (using `geos-3.3.3`).  Our shapely, `1.5.8dev0`, is compiled against `geos-3.4.2`.  The problem is that our shapely is not downloaded because `1.5.8` takes precedence over `1.5.8dev0`.

I will fix this.  Meanwhile do not merge this PR nor #214.

PR #219, opened 2015-05-05
Stable release of cf_units

ocefpaf:
@rsignell-usgs first stable release of cf_units.


rsignell-usgs:
Okay, I see now on sci-tools.  Awesome!

PR #220, opened 2015-05-05
Deprecate tappy

### LINKED ISSUE #174: Deprecate tappy in favor of Utide
https://github.com/wesleybowman/UTide

ocefpaf:
We already have Utide as a conda package.  Closes #174.


rsignell-usgs:
Is the functionality better in Utide than tappy, or are we deprecating tappy just because there is active development on Utide, and there isn't on tappy?

ocefpaf:
> Is the functionality better in Utide than tappy, or are we deprecating tappy just because there is active development on Utide, and there isn't on tappy?

Both.

If we ever get a new tappy release we can build that, but for now I want to remove it from being picked up by obvious-ci to speed builds up a little bit.  (Even when the recipe is not built it takes some time to check it and pass.)

Note that this does not remove the previously built binary.  Tappy will still be available on ioos channel.

PR #228, opened 2015-05-06
Testing basemap on windows

ocefpaf:
Trying to fix basemap for Windows.  If this does not work I will remove the recipe.  After all, the issue with the geos version was shapely and not basemap.


rsignell-usgs:
we don't need this now, right?

ocefpaf:
Nope!

PR #229, opened 2015-05-07
pyaxiom 0.0.9 released

kwilcox:


rsignell-usgs:
@ocefpaf , do we expect this to fail on mac because of the nco requirement?

kwilcox:
I commented out the `nco` requirement and it warns the user if they try to use a function that requires NCO that it needs to be installed.  https://github.com/axiom-data-science/pyaxiom/blob/master/requirements.txt.  We could (potentially) remove the requirement for the `nco-bindings` package, with the result being a limited set of functions without the user specifically installing `nco-bindings` manually.

ocefpaf:
@kwilcox you do not have to do that in the `requirements.txt`.  We can just added `# [not osx]` in front of the `nco` requirement in the `meta.yaml` file.  As long as it is an optional import (and I think it is) things will just fine.

The reason I did not do that before is because I thought `nco` was a low hanging fruit on Macs :unamused: 

@rsignell-usgs I am merging this and fixing the build.  `nco`, `nco-bindings`, and `pyaxiom` are not built on Mac due to a hack (https://travis-ci.org/ioos/conda-recipes/builds/61657224#L101) I used to get all the packages to build on OSX while migrating to the travis system.

PR #230, opened 2015-05-07
Fixed pyaxiom build for Windows and Mac

ocefpaf:
@kwilcox This complements PR #229.

Just some notes on conda packaging:

In conda we reset the build number every new version.  And we bump the build number when changing the package in the same version.  Suppose we get `nco` going for Mac we would create pyaxiom-0.0.9-**1**.

Also, the conda syntax to apply (or do not apply) specific changes to a certain platform are listed here:
http://conda.pydata.org/docs/build.html#preprocessing-selectors

Thanks for the PR!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
         - CONDA_PY=27  CONDA_NPY=19
 
 before_install:
-    - rm -rf nco nco-bindings pyaxiom
+    - rm -rf nco nco-bindings
     - brew update
 
 install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,7 +9,7 @@ env:
         - CONDA_PY=27  CONDA_NPY=19
 
 before_install:
-    - rm -rf nco nco-bindings pyaxiom
+    - rm -rf nco nco-bindings
```
ocefpaf: @rsignell-usgs now we have `pyaxiom` for Mac (and Windows) sorry for the misleading false positive in PR #229 due to my hackish way to avoid the build on travis.  This PR is safe to be merged.

rsignell-usgs:
okay, I get it now.

PR #232, opened 2015-05-08
Use 32 or 64 flags.

ocefpaf:
Just to make 32 and 64 builds a little bit easier.  I am not bumping the build number on purpose.  The 64 `geos` would not change and the 32 is already in the channel, which uses this PR and the docker image.

### DIFF
--- a/geos/build.sh
+++ b/geos/build.sh
@@ -1,7 +1,18 @@
 #!/bin/bash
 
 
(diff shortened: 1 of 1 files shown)


rsignell-usgs:
those changes make sense to me.

ocefpaf:
@pelson does this makes sense to you too?  If so I can send a PR to scitools-conda-recipes.

pelson:
:+1: - that would be useful to share onto conda-recipes-scitools

PR #234, opened 2015-05-08
Added bld.bat for Windows pyugrid.

ocefpaf:
Missed that `bld.bat` :flushed:


rsignell-usgs:
pyugrid for windows!  Woo hoo!

PR #235, opened 2015-05-11
Update utilities

ocefpaf:
Version 0.2.3

* `is_model` now assumes that all "GRID" features that have the word AVHRR are
  satellite data and not numerical models.


ocefpaf:
@rsignell-usgs this PR is good to merge.  (The other two are not, I am waiting for an official release to avoid using master.)

PR #240, opened 2015-05-15
Added django-extensions.

ocefpaf:
This PR adds the last missing sci-wms dependency listed in `requirements.txt`, `requirements-dev.txt`, and `requirements-prod.txt`.

@kwilcox let me know if there is anything else you need.  I would like to have all the dependencies on the IOOS channel to make it easier for future users.

### DIFF
--- a/argh/bld.bat
+++ b/argh/bld.bat
@@ -1,0 +1,1 @@
+"%PYTHON%" setup.py install --single-version-externally-managed --record record.txt
--- a/argh/build.sh
+++ b/argh/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install --single-version-externally-managed --record record.txt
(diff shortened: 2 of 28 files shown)


rsignell-usgs:
@kwilcox, ping?

kwilcox:
* django-grappelli
* django-typed-models
* dj-static

ocefpaf:
Thanks @kwilcox!

@rsignell-usgs I will add those here so hold off merging for a while...

rsignell-usgs:
@ocefpaf, ready now?

ocefpaf:
Not yet...

ocefpaf:
Phew!  Now it is!

Note to self: Some of the `django.<something>` return a `django.core.exceptions.ImproperlyConfigured`, but I think that is OK.  Only with a proper configured env this will go away.  (Maybe I should test that in the sci-wms metapackage...)

PR #246, opened 2015-05-28
Testing latest shapely.

ocefpaf:
**Do not merge!**

Just testing latest shapely...


ocefpaf:
We probably need to update the Windows patch.

ocefpaf:
Both Shapely and Pyresample (  #261 ) are tripping due to the lack of `stdint.h'` in Visual Studio.

`fatal error C1083: Cannot open include file: 'stdint.h': No such file or directory`

There are a few workarounds for this issues.  I will investigate which one serves us best.

rsignell-usgs:
Dang it.  What a pain!

ocefpaf:
Actually, after SciPy, I am positive that the default channel will provide
a decent shapely. Soon we will be deprecating ours.

PR #252, opened 2015-06-03
iris 1.8.1

ocefpaf:
# Bugfixes since v1.8.0:

- The PP loader now carefully handles floating point errors in date time conversions to hours.
- The handling fill values for lazy data loaded from NetCDF files is altered, such that the _FillValue set in the file is preserved through lazy operations.
- The risk that cube intersections could return incorrect results due to floating point tolerances is reduced.
- The new GRIB2 loading code is altered to enable the loading of various data representation templates; the data value unpacking is handled by the GRIB API.
- Saving cube collections to NetCDF, where multiple similar aux-factories exist within the cubes, is now carefully handled such that extra file variables are created where required in some cases.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,6 +24,7 @@ platform:
     - x64
 
 install:
+    - cmd: rmdir C:\cygwin /s /q
     - cmd: python bootstrap-obvious-ci-and-miniconda.py %CONDA_INSTALL_LOCN% %TARGET_ARCH% %CONDA_PY:~0,1% --without-obvci
     - cmd: set PATH=%PATH%;%CONDA_INSTALL_LOCN%;%CONDA_INSTALL_LOCN%\scripts
     - cmd: set PYTHONUNBUFFERED=1
--- a/bootstrap-obvious-ci-and-miniconda.py
+++ b/bootstrap-obvious-ci-and-miniconda.py
@@ -23,7 +23,7 @@ MINICONDA_URL_TEMPLATE = ('http://repo.continuum.io/miniconda/Miniconda{major_py
                           '{miniconda_version}-{OS}-{arch}.{ext}')
 
 
-def miniconda_url(target_system, target_arch, major_py_version, miniconda_version='3.7.0'):
+def miniconda_url(target_system, target_arch, major_py_version, miniconda_version='latest'):
     template_values = {'miniconda_version': miniconda_version}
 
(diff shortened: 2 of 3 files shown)


ocefpaf:
Latest AppVeyor\* broke our builds :unamused:

My understanding is that `Cygwin` is the default shell now and that is failing when conda-build calls git.

(See https://ci.appveyor.com/project/comtbot/conda-recipes/build/1.0.418/job/6lbtfkxb6o1hiq7k#L202)

I made a few tests with no avail.  I am reading the docs now to see what can be done, but it is hard to tests things without a Windows machine...

@pelson Are you experiencing this in scitools-conda-recipes?

\* But some of those updates will make our job easier in the neat future: http://www.appveyor.com/updates

PS: I am not sure why the the first git call is not from cygwin, just the calls from conda_build via `subprocess`.

rsignell-usgs:
@ocefpaf have you tried asking on appveyor forum? 
http://help.appveyor.com/search?q=conda&t=d&scope=all&category_id=
Might be worth a try...

ocefpaf:
[Done](http://help.appveyor.com/discussions/problems/2255-latest-appveyor-added-cygwin-and-that-broke-my-current-workflow).

I notice that conda seems to be using a mixed path syntax when building the `git clone` command:

```python
['C:\\cygwin\\bin\\git.exe', 'clone',
 '--recursive',
u'C:/conda/conda-bld/git_cache/__github.com_axiom-data-science_pyaxiom',
 'C:\\conda\\conda-bld\\work']
```

Maybe that is a "conda on Cygwin" bug...

ocefpaf:
@rsignell-usgs the workaround is to delete the `Cygwin` directory.  I will try to find a better fix, but this is OK to merge.

rsignell-usgs:
Well done sir!

pelson:
Looks like we have a bug somewhere. I'm a bit surprised that AppVeyor have turned on Cygwin by default... I think just killing their git might have been sufficient (to pass this part at least). It was calling ``'C:\\cygwin\\bin\\git.exe``.

ocefpaf:
It makes sense to delete just `git.exe`, no need for my :bomb: approach.

BTW: This is the AppVeyor thread about this: http://help.appveyor.com/discussions/problems/2255-latest-appveyor-added-cygwin-and-that-broke-my-current-workflow

PR #256, opened 2015-06-06
Deprecate rasterio.

ocefpaf:
`rasterio` became a first class citizen now!  The latest version is present for all archs (including Windows!) in the default channel.


ocefpaf:
Merging this one so our users to do get confused with the default `rasterio`.

PR #259, opened 2015-06-09
Adding tardis and updated utilities.

ocefpaf:
I factored `iris_utils` out of `utilities` and added it as a dependency.  Nothing changes if someone install installs `utilities`, but now people can use `iris_utils` without installing unnecessary modules.


ocefpaf:
Self-merging this... I need it for SECOORA.

PR #260, opened 2015-06-10
Update xray and adds some optional dependencies

ocefpaf:
This PR adds:
- `cyordereddict`
- `h5netcdf`

and updates `xray` to `0.5.0` using `h5netcdf`, `cyordereddict` and `dask`.

### DIFF
--- a/cyordereddict/bld.bat
+++ b/cyordereddict/bld.bat
@@ -1,0 +1,1 @@
+"%PYTHON%" setup.py install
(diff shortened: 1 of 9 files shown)


ocefpaf:
@rsignell-usgs xray made it to the default conda channel, but I do not see these optional dependencies (that make is faster, lighter, and meaner) in the default package.

rsignell-usgs:
Okay, let's keep ours then!

PR #261, opened 2015-06-10
Added Pyresample

ocefpaf:
Pyresample will help handle 2D coords in the iris wrapper module TARDIS.  (See #259.)


rsignell-usgs:
@ocefpaf , do we expect failure on windows?

ocefpaf:
Yes, but it is fixable. It cannot find `stdint.h`, probably because of AppVeyor's outdated Visual Studio.  I have a few ideas to fix this.  Lets leave this PR open until then.

ocefpaf:
I tried everything in [here](http://stackoverflow.com/questions/126279/c99-stdint-h-header-and-ms-visual-studio) with no avail.  I decided to drop `pykdtree` support on Windows and use scipy's KDTree backend instead for  `pyresample`.

PR #268, opened 2015-06-17
Update meta.yaml

jdhughes-usgs:
Updated package information


jdhughes-usgs:
Updated flopy package information

PR #270, opened 2015-06-17
Added siphon

ocefpaf:
# Siphon
- Siphon - A collection of Python utilities for retrieving data from Unidata data technologies.

An example from a mailing list:

```python
# Resolve the latest HRRR dataset
from siphon.catalog import TDSCatalog
latest_hrrr = TDSCatalog('
http://thredds-jumbo.unidata.ucar.edu/thredds/catalog/grib/HRRR/CONUS_3km/surface/latest.xml
')
hrrr_ds = list(latest_hrrr.datasets.values())[0]

# Set up access via NCSS
from siphon.ncss import NCSS
ncss = NCSS(hrrr_ds.access_urls['NetcdfSubset'])

# Create a query to ask for all times in netcdf4 format for
# the Temperature_surface variable, with a bounding box
query = ncss.query()
query.all_times().accept('netcdf4').variables('Temperature_surface')
query.lonlat_box(45, 30, -100, -90)

# Get the raw bytes and write to a file.
data = ncss.get_data_raw(query)
with open('test.nc', 'wb') as outf:
    outf.write(data)
```

### DIFF
--- a/protobuf/bld.bat
+++ b/protobuf/bld.bat
@@ -1,0 +1,1 @@
+"%PYTHON%" setup.py install --single-version-externally-managed --record record.txt
--- a/protobuf/build.sh
+++ b/protobuf/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install --single-version-externally-managed --record record.txt
--- a/protobuf/meta.yaml
+++ b/protobuf/meta.yaml
@@ -1,0 +1,33 @@
+package:
+    name: protobuf
+    version: "3.0.0a3"
+
+source:
+    fn: protobuf-3.0.0a3.tar.gz
+    url: https://pypi.python.org/packages/source/p/protobuf/protobuf-3.0.0a3.tar.gz
+    md5: 6674fa7452ebf066b767075db96a7ee0
+
+build:
+    preserve_egg_dir: True
+    number: 0
+
+requirements:
+    build:
+        - python
+        - setuptools
+    run:
+        - python
+        - setuptools
+
+test:
+    imports:
+        - google
+        - google.protobuf
+        - google.protobuf.compiler
+        - google.protobuf.internal
+        - google.protobuf.pyext
+
+about:
+    home: https://developers.google.com/protocol-buffers/
+    license: New BSD License
+    summary: 'Protocol Buffers'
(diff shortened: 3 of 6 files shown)


rsignell-usgs:
I'm not sure I understand the utility of this package.   Why would you get data via NCSS instead of OPeNDAP?

Are you using it?

ocefpaf:
I am experimenting with it for now.  Hold off the merge until we know that this is useful for a broader audience.

rsignell-usgs:
@ocefpaf , have you determined whether this is faster than OPeNDAP?  
Maybe that's the usefulness?

ocefpaf:
I haven't had time to play with this package yet.

Maybe I will cram that into my extra leap-second for today :wink: 

<sarcasm> The docs are amazing!!!</sarcasm>  See: http://siphon.readthedocs.org/en/stable/

PS: Based on the code example above I believe the usefulness is to explore the catalog directly.  Not sure though...

emiliom:
@ocefpaf, I can see that you probably were able to use the extra leap second to read the entire documentation ...

I haven't checked out siphon either, but it's in my indefinite to-do list to find a decent module to explore THREDDS catalogs. I haven't tried thredds_crawler either (and I see it's on the ioos channel), but I now wonder which one to invest time on. I would hope a package from unidata would be better in the long run ...

ocefpaf:
>  I can see that you probably were able to use the extra leap second to read the entire documentation ...

Twice actually :wink: 

Both are on my list too.  If I ever write about them I'll ping you for a review.

cehbrecht:
i was looking for a tool to navigate in thredds catalogs and retrieve metadata and access urls. First i tried the `thredds_crawler` ... easy to find on pypi ;) But the `thredds_crawler` does not only parse thredds catalog xml pages but also runs an extra query for each found dataset (with dataset_id). That puts quite some traffic on the thredds server and i have not seen a case where the dataset queries gives more information ...

Then i found the `threddsclient` from ScottWales ... it was half finished but more in the way i wanted it ... so i worked on that one: https://github.com/bird-house/threddsclient

After if have done this i found `siphon` ;) So, my hope is that `siphon` will provide the functionality i want ... but it is still work in progress. They already know about `threddsclient` and Scott seams also to have started to work on `siphon`.

PR #284, opened 2015-06-19
Bug in jasper release

ocefpaf:
Newer RHEL jasper had a lower build number than an older version.  That was causing all sort of issues with other packages that depended on jasper.


rsignell-usgs:
Is this an issue in general, that a newer package with a lower build number cause problems?
I thought each new version should build up from 0?

ocefpaf:
I used "version", but they are all the same version.

The higher build number was older, compiled in a modern Linux (Travis-CI), while the lower build number was newer (compiled against RHEL).  So conda was picking up the non-RHEL compat version.

Issue #245, opened 2015-05-25
nco-bindings


ocefpaf:
We first need nco and then we can build `nco-bindings` on OSX.

ocefpaf:
Fixed in #285

Issue #221, opened 2015-05-05
nco


ocefpaf:
@ChrisBarker-NOAA is this something that you might want to take a look?

- [X] nco
    - [X] nco-bindings
    - [X] pyaxiom

ChrisBarker-NOAA:
NCO -- yes

pyaxium -- that's the first I've heard of it -- but it looks like it's got some good stuff.

But when is the question -- not today ;-)

ChrisBarker-NOAA:
Actually, it will be at least a few days -. My Mac is dead :-)

CHB

On May 5, 2015, at 11:38 AM, Filipe <notifications@github.com> wrote:

@ChrisBarker-NOAA <https://github.com/ChrisBarker-NOAA> is this something
that you might want to take a look?

   -  nco
      -  nco-bindings
      -  pyaxiom

—
Reply to this email directly or view it on GitHub
<https://github.com/ioos/conda-recipes/issues/221>.

ocefpaf:
@ChrisBarker-NOAA don't worry.  `nco` is a nice package to have in the channel, but I guess that most OSX users already have it installed via other methods.

ocefpaf:
BTW we only need `nco`, `pyaxiom` and `nco-bindings` are pure python modules that depend on nco.

ocefpaf:
Fixed in #285

PR #285, opened 2015-06-20
Testing NCO on OSX

ocefpaf:
Mac OS X build hints (from http://nco.sourceforge.net/build_hints.shtml):

Mac OS X users report success building NCO without patches so long as they
- (1) Install the latest version of bison
- (2) Disable shared libraries and
- (3) Disable extended regular expressions.

 This is done with

```bash
./configure --disable-regex --disable-shared
```

Building NCO with shared libraries on Mac OS X on may be possible.  To try, first re-build and re-install
the netCDF library using `CFLAGS=-fno-common`.  Apparently netCDF, unlike most codes, does not set
this automatically.  Then build NCO normally.

Mac OS X ships with a broken version of the POSIX `cut` utility which causes many NCO regression
tests to appear to fail.  In fact the NCO commands succeed, but the `cut` output fails.  The workaround is
not to heed the regression tests, or to install a working `cut`, e.g., from the GNU `coreutils` package.

PS: A homebrew and a port recipe for reference:

https://github.com/Homebrew/homebrew-science/blob/master/nco.rb

https://trac.macports.org/browser/trunk/dports/science/nco/Portfile

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
-# MacOS builds.
-language: objective-c
+language: objective-c  # MacOS builds.
 
 env:
     global:
@@ -9,8 +8,12 @@ env:
         - CONDA_PY=27  CONDA_NPY=19
 
 before_install:
-    - rm -rf nco nco-bindings
     - brew update
+    # NCO.
+    - brew outdated gcc || brew upgrade gcc
+    - brew install gsl
+    # https://travis-ci.org/ioos/conda-recipes/builds/67641190
+    #- brew install antlr2
 
 install:
     # Download and run bootstrap.
--- a/nco/build.sh
+++ b/nco/build.sh
@@ -3,17 +3,25 @@
 export HAVE_NETCDF4_H=yes
 export NETCDF_ROOT=$PREFIX
 
-./configure \
-    HAVE_ANTLR=yes \
-    --prefix=$PREFIX \
-    --with-zlib=$PREFIX \
-    --disable-dependency-tracking \
-    --enable-netcdf4 \
-    --disable-static \
-    --disable-udunits \
-    --enable-udunits2 \
-    --disable-dap-opendap \
-    --enable-dap-netcdf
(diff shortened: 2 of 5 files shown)


ocefpaf:
I cannot build the docs with OSX default (old) texinfo.  The online solution seems to be installing from ports (`sudo port install texinfo`), but ports is not available in `Tavis-CI`.

Homebrew is, but somwhow I was unable to make that work.  My solution is to patch the Makefile and bypass the docs :unamused:

ocefpaf:
This PR is finally ready!

Note that I was unable to build with `antlr2` due to some wackiness with OSX default GCC.  

@ChrisBarker-NOAA you are the Mac expert :wink: do you have an idea of what we can do to solve this:

https://travis-ci.org/ioos/conda-recipes/builds/67641190

(Closes #245 and #221.)

ChrisBarker-NOAA:
Is this all set then? Or does it need more OS-X testing / debugging?

ocefpaf:
The package is available at IOOS channel now, but without `ncap2` support (no `antlr2`).

I could not figure out a way to make that work on OSX.  Note that [antlr2](https://github.com/ioos/conda-recipes/blob/master/.travis.yml#L16) is commented out, and [here](https://travis-ci.org/ioos/conda-recipes/builds/67641190) is the error if I enable it back.  My guess is some issue with clang vs gcc, but I have no idea how to sort that out without an actual Mac in front of me.

ChrisBarker-NOAA:
OK -- no time to check into it now, but maybe a new ticket specifically for that issue, and hopefully I can take a look later.

ChrisBarker-NOAA:
By the way, that looks a bit like it's gotten confused about 64 bit vs 32 bit -- maybe that helps?

ocefpaf:
> new ticket specifically for that issue

Done in #287

> looks a bit like it's gotten confused about 64 bit vs 32 bit 

I Google that and all I could find were regarding some flags difference for clang vs gcc.

PR #288, opened 2015-06-24
adding simplejson and ckanapi

rsignell-usgs:


ckanapi/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+  # If this is a new build for the same version, increment the build
+  # number. If you do not include this key, it defaults to 0.
+  # number: 1
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - setuptools
```
ocefpaf: duplicated

ckanapi/bld.bat
```diff
@@ -0,0 +1,8 @@
+"%PYTHON%" setup.py install
```
ocefpaf: When a package uses `setuptools` we can add `--single-version-externally-managed --record record.txt` here and below.  That makes the package more "portable."

ckanapi/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+    - docopt
+    - requests
+    - simplejson
+
+  run:
+    - python
+    - setuptools
+    - setuptools
```
ocefpaf: Duplicated again.

ckanapi/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+  # number: 1
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - setuptools
+    - docopt
+    - requests
+    - simplejson
```
ocefpaf: If you add `setuptools` we can add `--single-version-externally-managed --record record.txt` you can probably remove `docopt`, `requests`, and `simplejson` from the build step.

simplejson/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install
```
ocefpaf: Same comment as abouve about the `--single-version-externally-managed --record record.txt`

ckanapi/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+    #
+    # - ckanapi = ckanapi:main
+    #
+    # Would create an entry point called ckanapi that calls ckanapi.main()
+
+    - ckanapi=ckanapi.cli.main:main
+
+  # If this is a new build for the same version, increment the build
+  # number. If you do not include this key, it defaults to 0.
+  # number: 1
```
ocefpaf: Uncomment this and change to `0`

simplejson/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    # syntax is module:function.  For example
+    #
+    # - simplejson = simplejson:main
+    #
+    # Would create an entry point called simplejson that calls simplejson.main()
+
+
+  # If this is a new build for the same version, increment the build
+  # number. If you do not include this key, it defaults to 0.
+  # number: 1
```
ocefpaf: Uncomment this and change to `0`

ocefpaf:
@rsignell-usgs I just reviewed and travis is complaining about the lack of `mock` to run the tests.  Adding that to the run dependencies is not recommended because we won't need `mock`.  I would remove https://github.com/ioos/conda-recipes/pull/288/files#diff-5905a7c04e784b169dc52ffd8dac8a5eR52 instead.

Also, for clarity in future changes, I recommend removing all the "commented out boilerplate" that conda generates when creating the package files.

ckanapi/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+  md5: 2ab2508df730b0e7d84e13be74adc314
+
+build:
+    number: 0
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - docopt
```
ocefpaf: These are `run` dependencies and should be below.

PS: In my comment I mentioned `build` and not `run` :stuck_out_tongue_winking_eye:

simplejson/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+  imports:
+    - simplejson
+    - simplejson.tests
+
+about:
+  home: http://github.com/simplejson/simplejson
+  license: MIT License or Academic Free License (AFL)
+  summary: 'Simple, fast, extensible JSON encoder/decoder for Python'
+
+# See
```
ocefpaf: Left over boilerplate comments.

rsignell-usgs:
Do things like this work "as is" in conda or do we need to add more stuff to get the "speedups" to work?
https://github.com/simplejson/simplejson/blob/master/setup.py#L89

ocefpaf:
Setuptools should take care of that.  We can add an "import speedup" test to be sure the build works.

PR #290, opened 2015-06-24
Removed unnecessary build reqs.

ocefpaf:
Just some cleanups.  In theory the build should get slightly faster by not calling those dependencies twice.


ocefpaf:
Self merging... this one is "Mostly Harmless."

PR #297, opened 2015-07-07
Geospatial

ocefpaf:
Adds:
- descartes
- geopandas
- geojsonio.py

and some its dependencies.


ocefpaf:
@rsignell-usgs is probably heading to Austin now.  So I am self-merging this one...

PR #301, opened 2015-07-07
Add paegan-transport

daf:


paegan-transport/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+        - python
+        - setuptools
+    run:
+        - python
+        - paegan
+        - fiona
+        - requests
+        - redis
+        - geojson
+        - pytables
```
ocefpaf: @daf apparently you are missing `redis` here.  (It is available in conda.)

daf: note line 22...

daf: @ocefpaf i'm guessing a channel issue?

ocefpaf: It is available in the default channel.  Let me check it here...

paegan-transport/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+requirements:
+    build:
+        - python
+        - setuptools
+    run:
+        - python
+        - paegan
+        - fiona
+        - requests
+        - redis
```
ocefpaf: Change to `redis-py`

daf:
@ocefpaf using `redis-py` fixed travis and broke appveyor :(

ocefpaf:
A quick search binstar search does not show `redis-py` for Windows :unamused:

I am unsure is [this one](https://binstar.org/Trentonoliphant/redis/files) is the right redis due to the naming issue.  If redis is an optional dependency we can add a `# [not win]` in the `meta.yaml` file, but that will require some changes in the `paegan-transport`.

PS: Is redis a pure python module?  What are the dependencies?  How difficult would it be to create a Windows recipe?

daf:
If i'm reading this properly, it has no deps at all: https://github.com/andymccurdy/redis-py/blob/master/setup.py

I've no experience creating a Windows recipe but could try it.

ocefpaf:
> I've no experience creating a Windows recipe but could try it.

Just do the same thing as you did with pysgrid.  (Note that as long as you have a `bld.bat` you have a windows recipe.)

ocefpaf:
BTW, the README says:

> redis-py requires a running Redis server. See Redis's quickstart for installation instructions.

Could that server be the Windows roadblock?

daf:
I think it just means the library won't work without a running redis server to connect to (seems pretty self-evident to me).

ocefpaf:
- Is `redis-py` (#302) without [the redis server](http://redis.io/) enough for `paegan-transport` to work properly?

- Can `redis-py` be an optional dependency of  `paegan-transport`?

daf:
- Yes
- Also yes (should I change the recipe to indicate this, or change paegan-transport to late import redis stuff, or both?)

ocefpaf:
> Also yes (should I change the recipe to indicate this, or change paegan-transport to late import redis stuff, or both?

Both!  That is a much better solution then providing a windows package that might be broken.

daf:
Agreed 100%.

daf:
How does one specify an optional deps in the conda recipe?

ocefpaf:
`# [not win]` in front on it.

ocefpaf:
See more at http://conda.pydata.org/docs/building/meta-yaml.html#preprocessing-selectors

daf:
@ocefpaf good now?

paegan-transport/meta.yaml
```diff
@@ -0,0 +1,41 @@
+package:
+    name: paegan-transport
+    version: "1.0.3dev"
+
+source:
+  git_url: https://github.com/axiom-data-science/paegan-transport.git
+  git_tag: db8f957f5f379def417f93c1941d3f20dbbf16bb
```
ocefpaf: Loved it!  Avoid master as much as you can.  (Releases as better though :stuck_out_tongue_winking_eye:)

paegan-transport/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - requests
+        - redis-py  # [not win]
+        - geojson
+        - pytables
+
+test:
+    imports:
+        - paegan.transport
+        - paegan.transport.bathymetry
+        - paegan.transport.controllers # [not win]
```
ocefpaf: Just out of curiosity.  This is due to the lack of `redis-py` or something else?

ocefpaf:
> @ocefpaf good now?

It was good before.  It is perfect now!  Thanks!!

PR #302, opened 2015-07-07
Add redis-py bindings (for windows)

daf:
Builds a package named redis-py, needed for #301

### DIFF
--- a/redis-py/bld.bat
+++ b/redis-py/bld.bat
@@ -1,0 +1,1 @@
+"%PYTHON%" setup.py install --single-version-externally-managed  --record record.txt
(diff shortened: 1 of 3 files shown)


ocefpaf:
Will this work without a redis server?  If not this package is broken!

daf:
@ocefpaf, I don't understand the question.  This is a client library to connect to a redis server, why would I have to package a server as well?  What do you mean by "work without"?

ocefpaf:
The official conda `redis-py` package depends on `redis-2.6.9`.  Maybe they think that people will use this locally to create a local server and use `redis-py`, maybe `redis-py` needs `redis`... I don't know because I am not a `redis` user.

If you really do not need `redis` to get `redis-py` working, then I am puzzled why continuum does not provide the Windows package as well...

daf:
Closing in favor of marking redis an optional dependency of paegan-transport in #301 (which it is).

PR #303, opened 2015-07-09
bumping vispy to non-dev 0.4.0 release

rsignell-usgs:


vispy/meta.yaml
```diff
@@ -1,10 +1,10 @@
...
-    version: "0.4.0.dev"
+    version: "0.4.0"
 
 source:
     git_url: https://github.com/vispy/vispy.git
-    git_tag: master
+    git_tag: v0.4.0
 
 build:
     number: 0
```
ocefpaf: Bump the build number!

rsignell-usgs:
okay, bumped the build number from 0 to 1.  I wasn't sure I needed to do that: I thought conda would realize that 0.4.0 is an update to 0.4.0.dev

ocefpaf:
Oops, you are right!  I did not notice that you changed from master to a release!  My bad :flushed: 

Merging it.

rsignell-usgs:
Ack, the packages are there, but `conda install` thinks that "0.4.0.dev" is more up to date than "0.4.0".   I will delete the old ones.   We should remember this when we name dev versions.

ocefpaf:
My bad. I need to pay more attention to https://www.python.org/dev/peps/pep-0440/

rsignell-usgs:
@ocefpaf , dang it
```
binstar remove ioos/vispy/1.4.0.dev
```
isn't working.  Do you know the correct syntax?

rsignell-usgs:
Nevermind, I removed using the GUI on binstar

ocefpaf:
I was going to suggest that :grimacing:

PR #308, opened 2015-07-19
Update paegan to 1.1.3

ocefpaf:
No info... no changelog.

PR #313, opened 2015-07-24
binstar -> conda-server

ocefpaf:
**Do not merge this!**

Travis is failing with,

```python
File "/Users/travis/miniconda/lib/python2.7/site-packages/obvci/conda_tools/build_directory.py", line 15, in <module>
    from binstar_client.utils import get_binstar
ImportError: No module named binstar_client.utils
```

It seems that continuum moved/renamed things breaking Obvious-CI.  If this works we should fix Obvious-CI instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ install:
     # Config conda channels.
     - conda config --set show_channel_urls True
     - conda config --add channels http://conda.binstar.org/ioos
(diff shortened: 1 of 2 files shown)


ocefpaf:
This works see #308 for the test and https://github.com/pelson/Obvious-CI/pull/13 for the fix.

PR #317, opened 2015-07-25
deprecate_boost

### LINKED ISSUE #295: Deprecate boost
@rsignell-usgs  boost just made it to the default channel (all platforms) and I suggest we deprecate it to help the channel maintainers :wink:

ocefpaf:
Boost is available in the default channel now.

Closes #295


rsignell-usgs:
Thanks for dropping this...

PR #319, opened 2015-07-27
Unneded deps on sciwms dependencies

ayan-usgs:
This looks pretty complete, though there are a few things that aren't needed anymore.
* South
* Shapely
* Basemap

You don't need the debug-toolbar unless you're running a development machine, but that's not that critical. A user with server access could disable it at his/her discretion by modifying settings.

I've also taken this opportunity to make a pysgrid release.

### DIFF
--- a/sci-wms/meta.yaml
+++ b/sci-wms/meta.yaml
@@ -22,7 +22,7 @@ requirements:
         - pytz
         - djangorestframework
         - pyugrid
-        - pysgrid == 0.1.0
+        - pysgrid
         - pyaxiom
         - django-grappelli
         - django-typed-models
--- a/sci-wms/meta.yaml
+++ b/sci-wms/meta.yaml
@@ -22,7 +22,7 @@ requirements:
         - pytz
         - djangorestframework
         - pyugrid
-        #- pysgrid  # We need a release!
+        - pysgrid == 0.1.0
         - pyaxiom
         - django-grappelli
         - django-typed-models
@@ -30,10 +30,6 @@ requirements:
         - dj-static
         - pandas
         - django-autoslug
-        # No longer in the requirements.txt! Are these still used?
-        - shapely
-        - south
-        - basemap
         # requirements-prod.txt
         - gunicorn >=0.13.4
         - setproctitle  # [not win]


rsignell-usgs:
Syntax should be `pysgrid ==0.1.0`, not `pysgrid == 0.1.0`

sci-wms/meta.yaml
```diff
@@ -22,7 +22,7 @@ requirements:
         - pytz
         - djangorestframework
         - pyugrid
-        #- pysgrid  # We need a release!
+        - pysgrid == 0.1.0
```
ocefpaf: @ayan-usgs conda's syntax for pinning version is different from pip's (it sucks, I know...), but I prefer if we do not pin the version.

ayan-usgs:
@ocefpaf,

It makes me uncomfortable not knowing the version number, but whatever, I've removed it.

ocefpaf:
> Syntax should be pysgrid ==0.1.0, not pysgrid == 0.1.0

It used to be just `=`, maybe they fixed that to be compatible with `pip`

> It makes me uncomfortable not knowing the version number, but whatever, I've removed it.

Unless there is a strong reason to pin a version (like backwards incompatible API) we should let conda deal with the versions update.

ayan-usgs:
> Unless there is a strong reason to pin a version (like backwards incompatible API) we should let conda deal with the versions update.

I'm not a conda user, nor am I pretending to be. I've burned enough times by missing version numbers in documentation to know that it is valuable enough to warrant inclusion. As far as I'm concerned, such matters are of developer preference.

rsignell-usgs:
@ayan-usgs, conda users will have a better experience if the version numbers are *not* included, unless there is a reason, as @ocefpaf notes.    You could do `>=1.1.0`, but if you pin it as `==1.1.0`, then folks will downgraded if a newer version exists.

ayan-usgs:
@rsignell-usgs,

So basically if sci-wms becomes unmaintained, but pysgrid and pyugrid continue to be maintained have get new releases which may or may not be backwards compatible, then then conda is smart enough to know which version to be used? Or even in the scenario where backwards incapable changes are made to pysgrid or pyugrid or any other package, unless conda knows what happened, their releases would have to be synced with those of sciwms in order to prevent it from breaking. While most developers strive to keep APIs the same, that's not always possible. It adds maintenance overhead. If I recall correctly, we've already wasted some time figuring out that sciwms can't use Django 1.8... we would have saved time some mucking around if that version number was included. You'll notice it's there now. 

I would much prefer people be using virtual environments for running their apps rather than have everything living together where version conflicts of version number would be minimized to begin with.

I've already removed the version number, so I don't see the point of beating this dead horse any further. It'll be the convention I use when working with folks for WHOI.

rsignell-usgs:
@ayan-usgs , I certainly agree that if we know that if we *know* we need a version older or newer than a certain version, we should specify it.   But otherwise not.   Then in the future we discover that a newer version of something breaks it, we can modify the requirements.

ayan-usgs:
@rsignell-usgs,

I think it's just a cultural difference. I and my colleagues have been savaged many times by minute differences between our workstation environments, dev, QA, and production server environments, that it makes sense to have strong situational awareness of what one actually has installed. Commonality throughout our stacks is very important to maintaining and operating our apps.

That said, I recognize that not all scenarios require that sort of situational awareness. I just wanted to make clear that there is a legitimate use for version numbers. It seemed necessary to challenge what seems to a subjective opinion that it was _better_ that they're not there. Such value judgements all too often depend on what one is doing with a particular code base. I prefer to adhere to the existing convention within a code base rather than change it. In this case, I failed to adhere to the existing convention, and I deserved to be poked for it.

PR #333, opened 2015-07-31
Deprecate coveralls

ocefpaf:
Coveralls is no longer unused by any package in the IOOS channel.


rsignell-usgs:
You removed your coveralls?  ;-)

ocefpaf:
Yep.  It should be used only as a test dependency in CIs.

PR #336, opened 2015-07-31
Updated betamax to 0.5.0

ocefpaf:


rsignell-usgs:
What do we use betamax for?

ocefpaf:
The dependency chain is  of `betamax`+`betamax-matchers` -> `github3.py` -> `geojsonio`.

So the right question is  What do we use `geojsonio` for?

PR #345, opened 2015-07-31
deprecate_python-dateutil

ocefpaf:
Made it into the default channel.


rsignell-usgs:
nice to see these making it into defaults...

ocefpaf:
Yep!  I really want to see the IOOS channel to hosting ~10-20 domain specific packages rather than everything that we need that it is not in the default.

PR #357, opened 2015-07-31
Update flopy version

jdhughes-usgs:


ocefpaf:
@jdhughes-usgs pandas is used but it is not in the dependency list:

https://github.com/modflowpy/flopy/blob/master/setup.py#L34

Can you updated the recipe to install pandas too?

PS: Note that importing the module at build time (https://github.com/modflowpy/flopy/blob/master/setup.py#L8) forces all the dependencies to be available before they are actually needed.  Prefer a pattern like [this](https://github.com/pyoceans/python-ctd/blob/master/setup.py#L19-L30) to inject the version number in the `setup.py`.

ocefpaf:
@jdhughes-usgs there is no need for #358, we can re-start this PR once you re-issue the 3.2.2 tag.

jdhughes-usgs:
Reissued 3.2.2 tag. Still getting used to requests. Sorry.

ocefpaf:
> Reissued 3.2.2 tag. Still getting used to requests. Sorry.

No problem.  The CIs are doing their thing and the PR will be ready in a few minutes.

PR #358, opened 2015-07-31
Update meta.yaml

jdhughes-usgs:
Update flopy version. Previous meta.yaml update attemp had code with import pandas that was not being used. Commented out for now. As a result no need for pandas in setup.py.

### DIFF
--- a/flopy/meta.yaml
+++ b/flopy/meta.yaml
@@ -1,10 +1,10 @@
 package:
     name: flopy
-    version: "3.2.1"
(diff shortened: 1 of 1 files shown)


ocefpaf:
@jdhughes-usgs  the git_tag 3.2.1 seems to be missing.  I guess you deleted it while fixing the pandas dependency.

PS: No need for this PR is the recipe is unchanged.  We can re-start the CIs (travis and AppVeyor)

PR #373, opened 2015-08-12
Updated tardis to v0.1.0

ocefpaf:
Version 0.1.0

* Fixed `get_surface()` that demand all the z points.
* Fixed `get_nearest_water()` to loop over all the points until actually finds
  water.
* Added `get_nearest_series()` when finding wanter is not an issue. (Much faster
  than `get_nearest_water()`.

### DIFF
--- a/tardis/bld.bat
+++ b/tardis/bld.bat
@@ -1,2 +1,1 @@
 "%PYTHON%" setup.py install --single-version-externally-managed --record=record.txt
-if errorlevel 1 exit 1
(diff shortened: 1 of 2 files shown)


rsignell-usgs:
So `get_nearest_series` works on unstructured grids which have no land, I assume.

ocefpaf:
> So `get_nearest_series` works on unstructured grids which have no land, I assume.

That is the idea behind its conception... If it works is another story...

PR #377, opened 2015-08-17
Add unit_conversion library

kwilcox:


rsignell-usgs:
What functionality is in this package https://github.com/NOAA-ORR-ERD/PyNUCOS that is not in https://github.com/SciTools/cf_units?

kwilcox:
For https://github.com/NOAA-ORR-ERD/PyGnome support

kwilcox:
It is very targeted to the needs of GNOME: https://github.com/NOAA-ORR-ERD/PyNUCOS/blob/master/unit_conversion/unit_data.py

PR #380, opened 2015-08-17
Add awesome-slugify library

kwilcox:


kwilcox:
For https://github.com/NOAA-ORR-ERD/PyGnome support

PR #382, opened 2015-08-17
Renamed

ocefpaf:
@kwilcox one missing letter the exact same day I cannot see properly!  That is bad timing Kyle!!


ocefpaf:
@rsignell-usgs do not merge this yet. I just noticed they are all like this (`yml` instead of `yaml`).  Let me fix them all in this PR first.

ocefpaf:
In this PR:

- Fix recipes extension `yml` to `yaml` (conda is not that smart);
- Use `--single-version-externally-managed --record record.txt` when `setuptools` is available;
- Added missing `colander` dependency (`iso8601`).
- Added `build: number: 0` to all recipes to keep track of the build and version resolutions.
- Added summaries for `testfixtures` and `regex`.

@kwilcox there is a newer version regex, but I am guessing you need `2013.12.31`, right?

kwilcox:
@ocefpaf I don't believe so, just need to satisfy the https://github.com/dimka665/awesome-slugify requirements.  I used that version because I copy and pasted from another recipe!

ocefpaf:
OK.  I just asked because every now and then I update everything in the channel and if some version must be fixed I like to know to avoid breaking stuff.

PR #385, opened 2015-08-24
Updated Shapely to 1.5.11

ocefpaf:
1.5.11
-------------------
- Replace packaging.version with distutils.version

1.5.10 (2015-??-??)
-------------------
- Monkey patch affinity module by absolute reference (#299).

### DIFF
--- a/shapely/geos_c.win.patch
+++ b/shapely/geos_c.win.patch
@@ -1,40 +1,16 @@
-diff -Naur ../Shapely-1.5.8.orig/setup.py ./setup.py
---- ../Shapely-1.5.8.orig/setup.py	2015-04-29 17:27:19.000000000 -0300
-+++ ./setup.py	2015-05-05 11:48:46.836524002 -0300
-@@ -181,7 +181,7 @@
-         "shapely.speedups._speedups",
-         ["shapely/speedups/_speedups.c"],
-         libraries=libraries,
--        include_dirs=[get_config_var('INCLUDEDIR')],),
-+        include_dirs=[],),
- ]
-
- cmd_classes = setup_args.setdefault('cmdclass', {})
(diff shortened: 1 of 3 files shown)


ocefpaf:
@rsignell-usgs **Do not merge this yet!!!**

I am just testing the latest shapely.

ocefpaf:
@pelson should I send this to conda-recipes-scitools as well?  Having the latest Shapely available in our channels  prevents conda from downloading the default channel package (which is compiled geos-3.3.3 and breaks cartopy).

PS: Shapely+geos will be my first conda-forge packages!  Just need some time...

pelson:
> Having the latest Shapely available in our channels prevents conda from downloading the default channel package (which is compiled geos-3.3.3 and breaks cartopy).

Ok. We fix that with features...

cartopy/meta.yaml

```
build:
  track_features:
    - correct_GEOS_identification_in_shapely  # [osx]
```

shapely/meta.yaml
```
build:
  features:
    - correct_GEOS_identification_in_shapely  # [osx]
```

A bit of testing might mean we only need the features and track_features in the shapely recipe, and that would be enough...
I've not played with features much, and don't know how well obiviousci will deal with them at this point.

ocefpaf:
> I've not played with features much, and don't know how well obiviousci will deal with them at this point.

Neither have I and, to be honest, it would be better if we could push `geos-3.4.2` and the `shapely` patch to the default channel.  I prefer to a "fix this" than "fix my case."

pelson:
Ping @asmeurer / @ilanschnell - I don't believe you are packaging shapely correctly (it requires a patch, as per this repo), and you are shipping an old GEOS. Any chance you could fix this?

asmeurer:
@groutr

PR #388, opened 2015-08-27
Python 3 builds

ocefpaf:
Fingers crossed!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,15 +8,16 @@ environment:
   CMD_IN_ENV: "cmd /E:ON /V:ON /C obvci_appveyor_python_build_env.cmd"
 
   BINSTAR_TOKEN:
+    # Generated with "binstar auth --create --name comtbot -o ioos" and encrypted on ci.appveyor.com/tools/encrypt.
     secure: a0u2Qr8yHKQQHhFB+oWUg2tWeQYyUxHtFVk1ofwIPDaW9L5FrIC8fSns+Zbq0Kno
 
   matrix:
     - TARGET_ARCH: "x86"
       CONDA_PY: "27"
-      CONDA_NPY: "19"
+      PY_CONDITION: "\"python >2.7,<3\""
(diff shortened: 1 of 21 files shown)


ocefpaf:
@rsignell-usgs AppVeyor is failing while downloading stuff from PyPI.  I am tempted to merge this as is and fix it as we go.

appveyor.yml
```diff
@@ -8,15 +8,16 @@ environment:
...
 
   BINSTAR_TOKEN:
+    # Generated with "binstar auth --create --name comtbot -o ioos" and encrypted on ci.appveyor.com/tools/encrypt.
     secure: a0u2Qr8yHKQQHhFB+oWUg2tWeQYyUxHtFVk1ofwIPDaW9L5FrIC8fSns+Zbq0Kno
 
   matrix:
     - TARGET_ARCH: "x86"
       CONDA_PY: "27"
-      CONDA_NPY: "19"
+      PY_CONDITION: "\"python >2.7,<3\""
```
pelson: Do you know something I don't? I think you need ``>=2.7`` :wink:

cf_units/meta.yaml
```diff
@@ -11,7 +11,7 @@ build:
 
 requirements:
     build:
-        - python
+        - python <3
```
pelson: Apparently this is problematic. It means that conda doesn't bake the python version :cry: - I think I need to fix that in conda-build...

appveyor.yml
ocefpaf: Oops :flushed:

cf_units/meta.yaml
ocefpaf: To be safe we have to specify `<3` in build and run then?

pelson: Yes. And any future work to allow py3 support would need to be a new build number.

ocefpaf: Thanks!

PR #390, opened 2015-08-28
Deprecate pyproj

ocefpaf:
Pyproj is in the default channel.


ocefpaf:
Got to love Windows! `awesome-slugify` builds just fine for Python3 on Linux, but Windows...  A fix is on the way.  Merging this to get some of the Python 3.4 on the channel.

PR #396, opened 2015-08-28
proj_python_dep

ocefpaf:
Same as https://github.com/SciTools/conda-recipes-scitools/pull/95


proj4/bld.bat
```diff
@@ -1,8 +1,3 @@
-nmake -f makefile.vc INSTDIR=%PREFIX%
-if errorlevel 1 exit 1
+nmake /f makefile.vc
 
-nmake -f makefile.vc INSTDIR=%PREFIX% install-all
-if errorlevel 1 exit 1
-
-move %PREFIX%\bin\*.* %PREFIX%
```
pelson: You may have lost functionality here. Might want to check.

ocefpaf: Indeed.  I blindly copied and paste from [scitools](https://github.com/SciTools/conda-recipes-scitools/blob/master/proj4/bld.bat) and forgot why I have [this](https://github.com/ocefpaf/conda-recipes/commit/0e9de7a25ad18ffa960131720b62a72dfd42bd32) there.


For future reference we need this line to copy `cs2cs.exe`, `geod.exe`, `nad2bin.exe`,  `proj.exe`, `proj.dll` to conda's prefix.

PS: The default channel has this package now (the name is `proj.4` ), but I [guess](https://github.com/conda/conda-build/issues/562) it will not work for us us. I will try it later anyways.

PR #398, opened 2015-08-28
Python 3 pyke

### LINKED ISSUE #391: Test pyke and iris on Python3
I am unsure if `pyke` will work properly.  The package builds just fine,

http://anaconda.org/SciTools/pyke/1.1.1/download/linux-64/pyke-1.1.1-py34_0.tar.bz2

but there is a lot of Python2 specific code in there.

ocefpaf:
Closes #391

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,13 +18,13 @@ environment:
       CONDA_PY: "27"
       PY_CONDITION: "python >=2.7,<3"
     - TARGET_ARCH: "x86"
-      CONDA_PY: "3"
+      CONDA_PY: "34"
       PY_CONDITION: "python >=3.4"
     - TARGET_ARCH: "x64"
       CONDA_PY: "27"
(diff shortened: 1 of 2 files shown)


pelson:
This is definitely the right approach, but it is interesting that the conda recipe needs the context of which python version *before* the meta has been loaded. That suggests that tools like ObviousCI will never be able to load the meta and only then determine which Python versions it should be built against...
it may be that ObviousCI needs to run https://github.com/conda/conda-build/blob/master/conda_build/metadata.py#L272 on the distributions later on, but there are still some introspection issues which I need to think about.

Sorry for the long, possibly pointless, paragraph. :+1: on this change

ocefpaf:
> Sorry for the long, possibly pointless, paragraph.

Don't be.  I always learn a lot from your comments.

>  :+1: on this change

I am puzzled by AppVeyor's failure.  It works just fine locally :unamused:

PR #399, opened 2015-08-28
Fix Python <3 builds

ocefpaf:
Both `build` and `run` python version must be defined.

### DIFF
--- a/cartopy/meta.yaml
+++ b/cartopy/meta.yaml
@@ -10,7 +10,7 @@ source:
         - cartopy.win.patch  # [win]
 
 build:
-    number: 0
+    number: 1
 
 requirements:
     build:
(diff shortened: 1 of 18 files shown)


pelson:
I think this is the way to go. There is a downside of not having conda bake the version, and that is: as it stands, it is possible to install the exact same distribution with py2.6 and py2.7 - but the contents will have been put in a python2.7/site-packages folder - essentially, this means the distribution is really only 2.7.

ocefpaf:
IMHO I believe that a build python dependency should always be baked in the package.  Or are there any downsides that I am not seeing?

pelson:
> IMHO I believe that a build python dependency should always be baked in the package.

Completely.

However, conda-build doesn't do that if there is *any* restriction on the version of Python: 
https://github.com/conda/conda-build/blob/51a1a44584a50d9fe5b3ae5d9540f07766f84768/conda_build/metadata.py#L351
where strictness is defined by https://github.com/conda/conda/blob/master/conda/resolve.py#L117.
Essentially, if there is anything other than ``python`` as a build dependency, then I believe the condition you put in will override the baking behaviour.

I am very keen to address this issue - I think I know how to fix it in conda-build, but it may take a little bit of work to change the way of thinking for one of these "MetaData" instances...

PR #401, opened 2015-08-29
Fix regression in the package

ocefpaf:
See https://github.com/ioos/conda-recipes/pull/396/files#r38259958


pelson:
P.S. Thoroughly recommend adding a test for this in the recipe.

ocefpaf:
Done in #404. (I will do a better one later.)

PR #404, opened 2015-08-31
Proj4 test

ocefpaf:
See https://github.com/ioos/conda-recipes/pull/401


ocefpaf:
Note that if the command fails `echo $?` is still `0`! :scream:

At least we will know that `cs2cs`, `proj`, and etc are there...

PR #409, opened 2015-09-02
Bake python string

ocefpaf:
Using `<3` in both the `build` and `run` produced packages without the `py` version string!  In this PR we test specifying a more restrictive condition on the `build` only.

http://anaconda.org/IOOS/iris/1.8.1/download/linux-64/iris-1.8.1-np_1.tar.bz2

### DIFF
--- a/cf_units/meta.yaml
+++ b/cf_units/meta.yaml
@@ -7,13 +7,13 @@ source:
     git_tag: v0.1.0
 
 build:
-    number: 1
+    number: 2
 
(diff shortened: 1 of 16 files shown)


pelson:
Yes, I think you've got it right. I'm not sure what I was thinking with my previous comment in another ticket...

ocefpaf:
No problem.  It took me a while to notice that was happening.  I think we are OK now.

PR #410, opened 2015-09-02
adding pygc

rsignell-usgs:
pygc is a great circle calculation package, a requirement for modflow2netcdf

### DIFF
--- a/pygc/bld.bat
+++ b/pygc/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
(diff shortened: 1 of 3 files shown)


ocefpaf:
@kwilcox Was this code part of paegan?

kwilcox:
Yeah I split it off some time ago from https://github.com/axiom-data-science/paegan/blob/master/paegan/external/greatcircle.py but paegan hasn't been updated to use this instead of the built-in stuff.

ocefpaf:
Awesome.  I am all for small self contained modules.

pelson:
Interesting timing. I merged this today into cartopy: https://github.com/SciTools/cartopy/blob/master/lib/cartopy/geodesic/_geodesic.pyx

ocefpaf:
> Interesting timing. I merged this today into cartopy:

Which I was going to suggest a separate module but forgot to write it there...

pelson:
> Which I was going to suggest a separate module but forgot to write it there...

I completely buy the small packages thing, but there is a lot of machinery that you have to duplicate when dealing with external C deps (such as you do with Proj.4) which made it less appealing. The pure python implementation is cool though!

PR #420, opened 2015-09-07
Windrose

ocefpaf:
Windrose is finally a package!


scls19fr:
From https://github.com/scls19fr/windrose/issues/12#issuecomment-138393049

PR #421, opened 2015-09-08
Updated cf_units to v0.1.1

ocefpaf:
- Python 3 support


ocefpaf:
Note to self: `pyke` is still broken :unamused: 
That is not an issue right now because iris for py3k is not ready yet, but we need to fix this.

ocefpaf:
@kwilcox here is the py3k version for `cf_units`.

PR #430, opened 2015-09-14
Deprecate suds

ocefpaf:
Suds is no longer developed.  I am deprecating it,but bare in mind that the binaries are still available in the IOOS channel.


ocefpaf:
The failure is due to attention of Python 3.5 and has nothing to do with the deprecation.  I am merging this.  (And investigating on how to get all of our packages built for Python 3.5.)

PR #431, opened 2015-09-14
Added suds-jurko

ocefpaf:
Latest update on `suds-jurko` was 2015-07-27, so that is better than suds... And we get Python3 too.


ocefpaf:
Pinging @dharhas

PR #437, opened 2015-09-20
Deprecate obvious-ci

ocefpaf:
It made sense to package `obvious-ci` in the past when we were asking for third parties help to OSX.  now that the CIs are working and pulling it from http://conda.binstar.org/pelson/channel/development it does not make sense to keep packaging it here.


pelson:
Eventually we'll need a recipe in conda-forge rather than depending on my development channel, but there are a number of repos using it at the moment, so I'm likely to fix any issues pretty quickly :wink:

ocefpaf:
> Eventually we'll need a recipe in conda-forge

:+1:

PR #442, opened 2015-09-24
Fixed nn source and license

ocefpaf:
**DO NOT MERGE THIS:**  I am waiting for an official release.  If that does not happen than we can consider merging it.

This PR fixes the license and source for the code.

NB: The `triangle.[c,h]` license is non-free for commercial use! :ghost:

### DIFF
--- a/nn/build.sh
+++ b/nn/build.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+cd nn
+
 ./configure --prefix=$PREFIX
 make
 make tests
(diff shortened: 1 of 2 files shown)


phobson:
I think the git tag here should be changed to Dr. Sakov's release (`"v1.85.0`")

ocefpaf:
I am not merging this one yet. We need a new release that has https://github.com/sakov/nn-c/blob/master/nn/CHANGELOG#L1-L4

I am leaving it here just in case release takes to long and we need to build it from master.

phobson:
Cool. Glad to know we're both paying attention ;)

ocefpaf:
And thanks for the review!

ocefpaf:
It does not look like a new release will be issued soon...  Merging it against master then.

PR #448, opened 2015-09-24
Updated gridutils to 1.0

ocefpaf:
This PR fixes code source, code license, updated version, and add some extra metadata.

Changelog
--------------

- v. 0.54, 5 May 2015
  - Introduced gu_setquitfn() to set the quit function externally;  the previous (internal) version is used as default.
- v. 0.53, 3 Mar 2015
  - Corrected the header guard in gridnodes.h
- v. 0.52, 4 Jul 2013
  - Added a new constructor gridnodes_create2()
  - Using a saved errno in strerror() now


phobson:
:+1: looks good too. URLs and version are correct.

PR #450, opened 2015-09-24
Updated Shapely to 1.5.12

ocefpaf:
Changes
=======

1.5.12 (2015-08-27)
-------------------
- Remove configuration of root logger from libgeos.py (312).
- Skip test_fallbacks on Windows (308).
- Call setlocale(locale.LC_ALL, "") instead of resetlocale() on Windows when
  tearing down the locale test (308).
- Fix for Sphinx warnings (309).
- Addition of .cache, .idea, .pyd, .pdb to .gitignore (310).


ocefpaf:
Merging this before the default channel updates their shapely and we start getting the `cartopy/geos` issue again...

PR #454, opened 2015-09-24
Python3 pyproj

ocefpaf:
I tried using the conda `proj4 4.9.2` via `setup-proj.py` instead of the bundled proj.4 source.  There are a few errors, mostly some small precision in the test number.  However, that raises the question:  how OK are we using `pyproj` and an old bundled `proj.4`?


rsignell-usgs:
so `pyproj` has it's own version of the `proj.4` C source code?

ocefpaf:
Yep.

PR #436, opened 2015-09-19
Enable Python3.5

ocefpaf:
~~**DO NOT MERGE THIS !!!**~~

- [X] Removed `django-debug-toolbar` dependency on `django <1.8`.
- [X] Default channel does not package `dateutil` for py35.  Ours seems to be working so I changed `pynco` do use that instead.
- [X] Removed `mock` in `github3.py`. In py35 `mock` is part of the standard library.
- [X] Default channel does not package `bottleneck` for py35.  Since it is an optional `xray` dependency I just removed it from the build.

PR #466, opened 2015-09-29
Allow django-debug-toolbar to use Django 1.8

kwilcox:


ocefpaf:
Awesome!  Can we remove that restriction everywhere?

If so I can do that in #436 while I keep testing what are the blockers to Python 3.5.

kwilcox:
We should check to make sure that all of the individual dependencies support 1.8.  1.8 is an LTS release and everything on 1.7 will (well, should) be ported to 1.8.

PR #468, opened 2015-09-29
increment pyGDP version number

ayan-usgs:
Incremented pyGDP release number. Other dependencies look fine. Ran a build -- no problems were observed.


ocefpaf:
@ayan-usgs Does this version works with Python 3?

ayan-usgs:
> Does this version works with Python 3?

Unfortunately, pyGDP is not safe to run on Python 3, so the `<3` will remain in place for now. One day though, it'll work with Python 3.

ocefpaf:
Thanks for the PR @ayan-usgs!

PR #484, opened 2015-10-07
Pygridtools from Geosyntec, new version

phobson:
Hey -- I made my employer the owner of the pygridtools library and upped the git tag to v0.2.

I also changed the dependencies of pygridgen  to skip basemap in favor of pyproj (should significantly reduce the build size and time).

### DIFF
--- a/pygridgen/meta.yaml
+++ b/pygridgen/meta.yaml
@@ -18,7 +18,7 @@ requirements:
         - pandas
         - numpy
         - matplotlib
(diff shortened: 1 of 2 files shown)


ocefpaf:
The PR is OK.  Those failures are not related to `pygritools`.

I am merging this, but it will take a while for the packages to show up because I need to find a fix for those failures.

Thanks @phobson!

PR #485, opened 2015-10-07
Pin proj version

ocefpaf:
For some reason that escapes me `cartopy` fails to build on Windows with `proj 4.9.2`. 

In this PR I re-add `proj 4.9.1` just to issue a new release (`5`).  I had a broken release somewhere and it is easier to re-build it then to find which one and delete the binary.

I also pin `numpy` to avoid picking up `1.10.0` due to `ValueError: numpy.dtype has the wrong size, try recompiling`.  I guess we will have to avoid a little bit to use `numpy 1.10.0`.

Note that I will return the `proj 4.9.2` in another PR.  Both OSX and Linux cartopy seemed OK with the newer version.

### DIFF
--- a/cartopy/meta.yaml
+++ b/cartopy/meta.yaml
@@ -17,9 +17,9 @@ requirements:
         - python
         - setuptools
         - six
-        - numpy
+        - numpy >=1.8,<1.10
         - cython
-        - proj4
+        - proj4 ==4.9.1  # [win]
         - geos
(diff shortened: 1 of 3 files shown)


ocefpaf:
Almost there... Still need to fix cartopy for Python 3.4 on Windows!

ocefpaf:
The failure is just a time-out.  This PR is finally good to go :tada:

PR #487, opened 2015-10-08
Merged all requirements in one file

### LINKED ISSUE #486: Exact requirements for macos and win64 but not linux?
The wiki instructions for linux use the [ioos_req.txt](https://github.com/ioos/conda-recipes/blob/master/00_env_requirements/ioos/ioos_req.txt) file that has no package version specification. I've used this successfully on multiple linux machines (all Ubuntu, BTW). 

But the macosx and windows64 instructions both use an "exact" req file (eg, [ioos_req_exact_mac_osx.txt](https://github.com/ioos/conda-recipes/blob/master/00_env_requirements/ioos/ioos_req_exact_mac_osx.txt)) that specifies exact package versions. Does the non-versioned requirements file not work on mac and windows?

A colleague of mine is trying to follow the IOOS wiki instructions for the first time, on macosx, and ran into this error when creating the IOOS env:
```
Fetching package metadata: ..
Error: No packages found in current osx-64 channels matching: binstar 0.10.1 py27_2
```
She's trying a couple of alternatives, but in the meantime I wanted to look into this a bit.

Thanks!

ocefpaf:
This is tested on Linux only and the docs/wiki needs an update.

The preferred way to install the whole scientific stack from the IOOS channel is:

`conda env create environment.yml`

an alternative is,

 `conda create --yes -n IOOS --file requirements.txt python=2.7`

Note that everything install on Windows, but I am facing an issue with iris and udunits2 that I do not believe is related to the env file.  (See https://github.com/SciTools/conda-recipes-scitools/issues/115.)

This PR closes #486

I really appreciate some tests! (In case OSX and Windows users are following this repository.)


emiliom:
Cool, thanks @ocefpaf. FYI, I may be in a good position to provide testing of environment creation on both Windows & OSX, next week. I'm working with two people who are new to Conda and are installing it, including envs from the IOOS channel, on their respective computers (one on OSX, the other on both Ubuntu and Windows).

ocefpaf:
@emiliom I tested on Linux and Windows 32.

On Windows be aware of the drive [letter issue](https://github.com/SciTools/conda-recipes-scitools/issues/115#issuecomment-146823225)!

I could not test it on Windows 64 and OSX.


Somethings that people should know when using the env files:

- The env installs virtually everything in the channel.  That might be too much for some application.
- `pynco` and `nco` are not installed because they are not available on Windows.  I wanted a single file that worked on Windows, Linux and OSX.  Users that need `nco` and `pynco` must install them afterwards.
- Nothing is versioned.  The env file is an example and users are encourage to edit them.  For exmple pinning some version down to guarantee reproducible envs, renaming the env name, adding more channels, adding removing packages, etc.
- The python version is not pinned, but since we still have some packages that are only available on python 27 users will get python 27 from that env.
- Some packages are listed even though they are dependencies for other packages.  I choose to be verbose and explicit rather than pushing a compact env file.  That way people can read the file and have an idea of what they will get.

rsignell-usgs:
On my win64 machine, I did:
```
conda env create environment.yml
```
and all the notebooks I've tried work.   Nice job!

So now that I have this environment, is there a way to say "upgrade everything"?

ocefpaf:
> So now that I have this environment, is there a way to say "upgrade everything"?

Not sure what you mean by "upgrade everything," but conda has,

```
conda update --all
```

to try and get the latest versions.

**CAVEAT** The latest numpy (1.10.0) might be a problem with some packages.  Try to downgrade numpy to 1.9 until I re-build everything against the new one.

PR #489, opened 2015-10-09
Binder badge

ocefpaf:
Added the binder badge so users can test the binaries here in an online interactive session of Jupyter Notebooks.


pelson:
**Way** cool @ocefpaf & @rsignell-usgs. Did you try changing the line ``%matplotlib inline`` for ``%matplotlib notebook`` - I thoroughly recommend you do :wink:

ocefpaf:
I write `inline` due to habit.  You are right. We should use `notebook`.

PR #490, opened 2015-10-09
Update notebooks

ocefpaf:
I updated the wiki and added ~~new~~ old example notebooks.


emiliom:
Awesome/Otimo, @ocefpaf! Looks great. Not that I read every page; I did look at all 4 notebooks (including the index nb).

ocefpaf:
The instructions are slightly simplified now.

Those notebooks can be run online [1] in the same environment we will get using the file [environment.yml](https://github.com/ioos/conda-recipes/blob/master/environment.yml).  So that is a quick way to test if things are working.

[1] http://mybinder.org/repo/ioos/conda-recipes

emiliom:
Nice! I noticed your mybinder comments earlier, but had not seen it in action until now.

Issue #491, opened 2015-10-09
Proj4 package name


ocefpaf:
The anaconda default channel names [Proj.4](https://trac.osgeo.org/proj/) as `proj.4` and we name it `proj4`.

Since we want to make use of the default channel as much as possible we should consider renaming ours, and eventually start using the default channel, to `proj.4`.

That will fix a few issues like having two proj4 packages installed when mixing default channel with our custom channels.

@pelson  What do you think?

pelson:
Yep, let's give it a shot! :+1:

Issue #413, opened 2015-09-04
Proj4 on win64 issues


rsignell-usgs:
I'm having trouble creating new environments with proj4 that work on win64.   

This environment doesn't work:
```
[ioos] C:\programs\anaconda-64>conda list proj
# packages in environment at c:\programs\anaconda-64\envs\ioos:
#
proj.4                    4.9.1                         0    defaults
proj4                     4.9.1                    py27_4    ioos
pyproj                    1.9.4                    py27_1    defaults
```
In another environment,  I uninstalled pyproj and proj4, then forced the install of proj4 from the ioos channel, and then tried to install pyproj from the ioos channel, but I couldn't get the ioos channel version because it has the identical version and build number as the default channel:
```
[pygdp] C:\programs\anaconda-64>conda list proj
# packages in environment at c:\programs\anaconda-64\envs\pygdp:
#
proj4                     4.9.1                    py27_4    ioos

[pygdp] C:\programs\anaconda-64>conda install pyproj
Fetching package metadata: ........
Solving package specifications: .
Package plan for installation in environment c:\programs\anaconda-64\envs\pygdp:


The following NEW packages will be INSTALLED:

    proj.4: 4.9.1-0      defaults
    pyproj: 1.9.4-py27_1 defaults

Proceed ([y]/n)? n
```
I then thought I could just bump the build number on the pyproj recipe, but found that it has been deprecated from the ioos channel.  

Luckily I still have one environment (I can't remember when I created it) that still works, with both proj4 and pyproj from the ioos channel:
```
[flopy] c:\Users\rsignell\Documents\GitHub>conda list proj
# packages in environment at c:\programs\anaconda-64\envs\flopy:
#
proj4                     4.9.1                    py27_4    ioos
pyproj                    1.9.4                np19py27_1    https://conda.binstar.org/ioos/win-64/pyproj-1.9.4-np19py27_1.tar.bz2
```

I'm supposed to help python windows users install working conda environments in a few hours, but I have no idea how to recreate my working environment!   :sob:

ocefpaf:
Did you try uninstalling ioos proj4 and use all from de default channel?  I
had hope to deprecate proj too.

I'm having trouble creating new environments with proj4 that work on win64.

This environment doesn't work:

[ioos] C:\programs\anaconda-64>conda list proj
# packages in environment at c:\programs\anaconda-64\envs\ioos:
#
proj.4                    4.9.1                         0    defaults
[...]

PR #500, opened 2015-10-10
Rename proj4 to proj.4

ocefpaf:
The default channel uses the name `proj.4` instead of `proj4` (see https://github.com/ioos/conda-recipes/issues/413#issue-104850388).  That results in two versions of Proj.4 being installed depending on the package the users are trying to install.  I am renaming ours to reflect the same names found in the default channel in order to avoid that.

Note that we are using `proj.4 4.9.2` everywhere but cartopy on Windows.

### DIFF
--- a/cartopy/meta.yaml
+++ b/cartopy/meta.yaml
@@ -10,17 +10,16 @@ source:
         - cartopy.win.patch  # [win]
 
 build:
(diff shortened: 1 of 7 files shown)


ocefpaf:
Closes #491

PR #528, opened 2015-10-21
Downgrade conda build

ocefpaf:
Latest conda-build introduced a [bug](https://travis-ci.org/ioos/conda-recipes/builds/85940423#L2326) that I believe is already fixed [here](https://github.com/conda/conda-build/commit/b758f1b1ac0c3efbb4d5cc486ab54270595356b4), but we need to downgrade conda-build until Continuum issues a new release.

Ping @pelson in case you get hit by this.


pelson:
Thanks @ocefpaf

PR #533, opened 2015-10-24
Latest netCDF4-python compiled with Cython

ocefpaf:
This PR fixes two problems with the default channel netCDF4-python

- compiles it with cython so we can use the `.filepath()` method
- Updates to 1.2.1 (the default channel is stuck at `1.1.9`)

```
 version 1.2.1 (tag v1.2.1rel)
 =============================
 * add the capability to slice variables with unsorted integer sequences,
   or integer sequences with duplicates (issue #467). This was done
   by converting boolean array slices to integer array slices internally,
   instead of the other way around.
 * raise TypeError if masked array assigned to a VLEN str variable slice
   (issue #464).
 * Ellipsis now can be used with scalar VLEN str variables (issue #458).
   Slicing of scalar VLEN (non-str) variables now works.
 * Allow non-positive reference years in non-real-world calendars
   (issue #442).

 version 1.2.0 (tag v1.2.0rel)
 =============================
 * Fixes to setup.py for building on windows (issue #460).
 * warnings now issued if file being read contains unsupported
   variables or data types (they were previously being silently
   skipped).
 * added 'get_variables_by_attributes' method (issue #454).
 * check for 'units' attribute in date2index (issue #453).
 * added support for enum types (issue #452).
 * added 'isopen' Dataset method (issue #450).
 * raise ValueError if year 0 or negative year used in time units string.
   The year 0 does not exist in the Julian and Gregorian
   calendars (issue #442).
```

### DIFF
--- a/netcdf4/build.sh
+++ b/netcdf4/build.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+SETUPCFG=$SRC_DIR\setup.cfg
+
+echo "[options]" > $SETUPCFG
+echo "use_cython=True" >> $SETUPCFG
+echo "[directories]" >> $SETUPCFG
+echo "netCDF4_dir = $PREFIX" >> $SETUPCFG
+
+${PYTHON} setup.py install --single-version-externally-managed --record record.txt
--- a/netcdf4/meta.yaml
+++ b/netcdf4/meta.yaml
@@ -1,0 +1,43 @@
+package:
+    name: netcdf4
+    version: "1.2.1"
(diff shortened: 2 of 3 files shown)


ocefpaf:
I failed miserably trying to get this to build on Windows.  I even tried re-building `hdf5` and `libnetcdf` from scratch, but netCDF4-python kept complaining about a missing `DLL`.

For now Windows will keep downloading an outdated netCDF4-python without Cython from the default channel.

PR #537, opened 2015-10-26
Bump thredds_crawler to 1.1.0 and use PyPI for source

kwilcox:
1.1.0 adds Python 3.4 support

### DIFF
--- a/thredds_crawler/meta.yaml
+++ b/thredds_crawler/meta.yaml
@@ -1,13 +1,14 @@
 package:
     name: thredds_crawler
-    version: "1.0.0"
+    version: "1.1.0"
 
 source:
-    git_url: https://github.com/asascience-open/thredds_crawler.git
-    git_tag: 1.0.0
+    fn: thredds_crawler-1.1.0.tar.gz
+    url: https://pypi.python.org/packages/source/t/thredds_crawler/thredds_crawler-1.1.0.tar.gz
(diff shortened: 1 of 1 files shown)


ocefpaf:
@kwilcox the latest Obvious-CI broke the way we do things here.  I will try to get that fixed.  Let's hold this PR for a while until then.

thredds_crawler/meta.yaml
```diff
@@ -19,6 +20,7 @@ requirements:
         - python
         - requests
         - lxml
+        - netCDF4
```
ocefpaf: Conda should not care, but this package is lower case in the default channel.

kwilcox: I'll fix. Question... is this the right place to put `netcdf4`? It's an optional dependency for one specific part of the code (and one test requires it).

ocefpaf: Optional dependencies are up to the packager and I like to add them all.
So that is good. Thanks!

However, we can probably drop the `requests` and `lxml` at the build time.  But l will check that later, it depends if the code import itself to get the version.

ocefpaf:
Only the Windows build for Python 3 are affected by https://github.com/pelson/Obvious-CI/issues/39.

So I am merging this.

kwilcox:
Thanks!

PR #538, opened 2015-10-26
Manual build string

ocefpaf:
Manually create build string.


ocefpaf:
@kwilcox with this change the Python 3 Windows version of  #537 should appear in the channel.

Issue #534, opened 2015-10-26
Possibility to also have a pyproj 1.9.3 version available?


jorisvandenbossche:
To use `pyproj` on Windows (64bit) and python 3.4, I need an older version (1.9.3), as all available 1.9.4 builds are not working (both in ioos as default conda channel). See the issue I reported here: https://github.com/jswhit/pyproj/issues/17

Current workaround I use it to install `pyproj` 1.9.3 from the JimInCO channel, which works, but is still one more complexity in the instructions I have to give to install geopandas on Windows. Would it be possible to also have this version available in the ioos channel? 

BTW, for the rest the ioos channel is really a life saver if you want to install these packages on windows! Thanks for that

ocefpaf:
Hi @jorisvandenbossche,

The IOOS channel does not package pyproj for Python 3 and Windows.  I tried an failed miserably to circumvent a known bug\*. (See https://github.com/ioos/conda-recipes/pull/480.)

The one you are getting is probably from the default channel. You can verify that by setting the **show channels** option and then using **conda list** like this:

```bash
conda config --set show_channel_urls True
conda list
```

\* The IOOS builds are automatic via AppVeyor, so it is complicated to patch `distutils/msvc9compiler.py` to fix this issue.

jorisvandenbossche:
Ah, thanks for the explanation! (that is one of the downsides of conda, it is not always clear where something is coming from when installing from a channel)

ocefpaf:
Indeed the `conda config --set show_channel_urls True` should be the default behavior.

I will leave this issue open.  I do plan to fix pyproj build for Win+py3k on day.

ocefpaf:
Fixe in #539

pelson:
> Indeed the conda config --set show_channel_urls True should be the default behavior.

How about a PR to conda/conda? https://github.com/conda/conda/pull/1771 :wink:

saulomeirelles:
Hi guys,

I just got in the same trouble as @jorisvandenbossche but using Python 2.7 on Windows 64-bits. Installing an older version from JimInCO channel did not work for me.

That happened after updating IOOS packages. I am just wondering if there is any progress on this issue. Would you mind informing me?

BTW, Cartopy is also behaving strangely with unexpected "OverFlow Error".

Thanks!

ocefpaf:
@saulomeirelles can you send us the results of the following commands?

```bash
conda config --set show_channel_urls True
conda list
```

[...]

PR #539, opened 2015-10-26
Test

ocefpaf:
Fix for #534 

@jorisvandenbossche can you test the binaries once the packages lands on the channel?  I don't have a Win64 machine.

### DIFF
--- a/pyproj/meta.yaml
+++ b/pyproj/meta.yaml
@@ -3,16 +3,15 @@ package:
     version: "1.9.4"
 
 source:
-    fn: pyproj-1.9.4.tar.gz
-    url: https://pypi.python.org/packages/source/p/pyproj/pyproj-1.9.4.tar.gz
-    md5: b0a060465b117dd6c1b37e14f22382aa
+    git_url: https://github.com/jswhit/pyproj.git
+    git_tag: master
 
 build:
     number: 2
 
 requirements:
     build:
-        - python >=2.7,<3
+        - python
     run:
         - python
         - numpy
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Yes, it works now for windows/64bit/python3 !
Thanks

ocefpaf:
Awesome!  Thanks for testing.

jorisvandenbossche:
Out of curiosity: what did fix this issue? Was it a bug in pyproj that is fixed by using master, or was it an issue in the packaging? (can the issue I reported to pyproj be closed? https://github.com/jswhit/pyproj/issues/17)

ocefpaf:
We have two different packing issues there:

- The default channel compiles pyproj against their version of proj.4 rather than the custom source shipped with pyproj's source.  That is the reason why the default channel is missing some features. (I added a [test](https://github.com/ocefpaf/conda-recipes/blob/5660cffc332b24ddbc4719d9e8e88c48364939a3/pyproj/run_test.py#L4-L5) to ensure we will always have the EPSG data in our package.)

- Pyproj 1.9.4 did not use setuptools and was hitting a bug in `distutils` that Python folks did not considered a [bug](https://bugs.python.org/issue4431).

A long time ago I tried to fix that in pyproj by adding setuptools, but my PR never moved [forward](https://github.com/jswhit/pyproj/pull/2). Luckily someone else did [this](https://github.com/jswhit/pyproj/pull/21) properly and that version of the PR got merged.

jorisvandenbossche:
Thanks for the explanation!

Issue #507, opened 2015-10-15
Numpy C-API


ocefpaf:
A list of packages that depends on the numpy ABI and require the `numpy x.x` syntax

- [X] cartopy
- [X] ecmwf_grib
- [X] mo_pack
- [X] shapely
- [X] iris
- [X] netcdf4
- [X] octant
- [X] gdal
- [X] fiona
- [X] rasterio

pelson:
Shapely?

ocefpaf:
Not sure if baking the build string will help.  The [conda's docs says](http://conda.pydata.org/docs/spec.html)

> The build string is used to differentiate builds of packages with otherwise the same name and version, e.g. a build with other dependencies (like Python 3.4 instead of Python 2.7), a bug fix in the build process, or some different optional dependencies (MKL vs. ATLAS linkage), etc. Nothing in conda actually inspects the build string—strings such as np18py34_1 are only designed for human readability, but are never parsed by conda.

UPDATE: However, it seems that the statement above is not true.  So I am re-adding the numpy build strings to the IOOS channel.

ocefpaf:
Outdated.

PR #540, opened 2015-10-29
Skip windows build

ocefpaf:
Not ready yet...


.travis.yml
```diff
@@ -37,14 +37,14 @@ before_install:
...
     - python bootstrap-obvious-ci-and-miniconda.py ~/miniconda x64 3 --without-obvci && source ~/miniconda/bin/activate root
     - conda config --set show_channel_urls True
     - conda install -c http://conda.binstar.org/pelson/channel/development --yes --quiet obvious-ci
     - conda config --add channels ioos
     - obvci_install_conda_build_tools.py
-    - conda install --yes conda-build==1.18.0
 
 script:
-    - obvci_conda_build_dir.py ./ ioos --channel main --build-condition "python >=2.7,<3|==3.4"
+    - obvci_conda_build_dir.py ./ SciTools --channel main --build-condition "numpy >=1.9" "python >=2.7,<3|>=3.4,<3.5"
```
pelson: Woops.

ocefpaf: OhOh!! Copy-and-paste :hankey:

No biggie.  This PR is far from ready :wink:

compliance-checker/meta.yaml
```diff
@@ -26,7 +22,7 @@ requirements:
         - requests
         - python-dateutil
     run:
-        - python >=2.7,<3
```
pelson: You've lost the fact that this is py2 only?

iris/meta.yaml
```diff
@@ -20,18 +16,22 @@ requirements:
         - biggus
         - cartopy
         - netcdf4
+        - numpy x.x
         - udunits2
         - pyke
         - setuptools
         - libmo_unpack  # [not win]
     run:
-        - python >=2.7,<3
```
pelson: Same as above. Is this in the build requirements in stead perhaps?

compliance-checker/meta.yaml
ocefpaf: But I thought that with the new modification we should constraint the python version at the `build` filed only.

ocefpaf:
Some lessons learned from this PR:

- Somewhere between `conda-build==1.18.0` and the latest `conda` version `conda-build` started choking on empty fields. Since we need the latest `conda` for this PR we also need to remove the empty fields. One might argue that we must to fix `conda-build` instead. However, that bug forces us to write cleaner `meta.yaml` :wink: (Changes: no longer downgrade `conda-build` and removed empty fields.)

- For some weird reason I could not get this PR working with miniconda3 on Travis-CI (OSX). That is not a problem because the encoding issue is finally fixed upstream. We are OK using miniconda2.

- In this PR I re-introduced the numpy build strings. (See https://github.com/ioos/conda-recipes/issues/507)

compliance-checker/meta.yaml
pelson: Ah. I see. My bad.

ocefpaf:
The AppVeyor failure is just a time-out.  It seems that this PR is finally good to go :tada: 

Thanks @pelson for the new goodies in `obvious-ci`!

PR #544, opened 2015-10-30
Added some tests

ocefpaf:
ping @kwilcox 

- Test `test_index()` and `test_property()` are seg-faulting because of the pickle!
- The `test_stream_input()` is failing with the assertion error with all `libspatialindex` versions I could find!
- And `test_non_stream_input()` is the same test data as `test_stream_input()`, but without the generator syntax.  This one passes :tada:


ocefpaf:
Note to self: Nothing gets build here because we don't need a new build.  I just want these tests to be there to check future changes.

PR #545, opened 2015-10-30
np string for cf_units

ocefpaf:
Numpy string for cf_units

### DIFF
--- a/cf_units/meta.yaml
+++ b/cf_units/meta.yaml
@@ -8,17 +8,18 @@ source:
     md5: 6dcaa762fbbb790684a2a918b5f34f5d
 
 build:
-    number: 0
+    number: 1
 
 requirements:
     build:
         - python
+        - numpy x.x
     run:
         - python
         - six
         - udunits2
         - netcdf4
-        - numpy
+        - numpy x.x
 
 test:
     imports:


pelson:
How come? I don't think there is a C api dependency is there?

ocefpaf:
Nope.  But we had an issue in compliace-checker where `cf_units` was downgrading numpy and breaking netCDF4-python.

I guess that now that we are using the np string again in ioos the problem should be gone.  I will experiment with this and remove the unnecessary `numpy x.x `

ocefpaf:
I read the [compliance-checker](https://travis-ci.org/ioos/compliance-checker/builds/87692366) failure more carefully and found out that the problem was an old version of `cf_units`, from the time that conda added the numpy string by default, that was downgrading numpy and breaking netCDF4-python.

So it makes sense to revert this PR and delete the binaries.

Thanks @pelson!

emiliom:
@ocefpaf, apologies for using this venue for asking a more general question. You said:
> I guess that now that we are using the np string again in ioos

I thought you had moved *away* from using np strings fairly recently (~ 2 months ago?), to follow what was being done by conda. Has that decision been reversed in the ioos recipes/channel?

ocefpaf:
That is a valid question!

I deleted the **empty** numpy string `np{nover}` packages\*.  That was an artifact due to a bug in conda-build when they decided to **stop** baking the numpy build string by default.  They fixed the bug at some point and we were left with several files with the `np` for no reason.

However, for longer than two months I was not baking the numpy string in **any** package, even those that depended on the numpy ABI.  I was doing that because conda docs mentioned they did nothing with the build string.  That is clearly not true.  Conda does take the build string in consideration!  So I decided to add the build string back, but only for the packages that depend on the numpy ABI. (See the list in #507.)

\* The reason behind the removal is to reduce the dependency resolve time.  Conda was choking on the huge number of choices for the same package.  For example: `gsw-np19py27_0`, `gsw-nppy27_0`, and `gsw-py27_0` **are** the same package and only the last one should stay in the channel (`gsw` does not depend on numpy's ABI and works with numpy version 1.7+).

emiliom:
Thanks. 

> The reason behind the removal is to reduce the dependency resolve time.

Yeah, you did explain this in response to another related, out-of-the-blue question I asked you.

PR #549, opened 2015-10-31
Added conda-execute

ocefpaf:
@pelson I wanted to avoid the duplication, but it seems that conda is forcing us in the wrong direction:

```shell
> anaconda copy conda-forge/conda-execute/0.4.0 --to-owner ioos
Using anaconda-server api site https://api.anaconda.org
[ServerError] Copying anaconda packages temporarily disabled.
This issue is being tracked here:
https://github.com/Anaconda-Server/docs.anaconda.org/issues/119 
```

At least it is "temporarily."

### DIFF
--- a/conda-execute/bld.bat
+++ b/conda-execute/bld.bat
@@ -1,0 +1,2 @@
+"%PYTHON%" setup.py install --single-version-externally-managed  --record record.txt
+if errorlevel 1 exit 1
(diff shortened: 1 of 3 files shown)


pelson:
Cool. No problems. I'm experimenting with allowing this repo to share recipes from a canonical source (see #554) - this would be one to consider after that is working as expected/desired.

PR #553, opened 2015-11-01
Test VC string bake

ocefpaf:
**DO NOT MERGE THIS!!!**

Just testing https://github.com/conda/conda-build/issues/562#issuecomment-152752047.


pelson:
Sounds like you can just depend on Python at build time?

ocefpaf:
At the moment we do add the python dependency at run time and that seems to work.

I just notice that the default channel started baking VC versions and I do not want our users to think they are getting the "wrong package" when they install our version just because of that.

However, I am not merging this.  I just wanted to check how `features` work.

pelson:
> However, I am not merging this. I just wanted to check how features work.

Did it not work as expected? I've not really tested obvious-ci against features, but I suspect it wouldn't take much work...

ocefpaf:
> Did it not work as expected?

AppVeyor failed with conda's most helpful error message:

```
Error: Unsatisfiable package specifications. Generating hint:

Hint: 'proj.4' has unsatisfiable dependencies (see 'conda info proj.4')
```

> I've not really tested obvious-ci against features, but I suspect it wouldn't take much work...

Not sure if it is worth it...

msarahan:
Yes, this is unfortunate.  If any dependency anywhere in the tree of dependencies hasn't been built with features, then this kind of thing can happen.  It is an extreme headache.  I will be pushing internally to make the error message more informative.

IMHO, features are well-worth adopting, because lining up the VC versions and keeping them consistent will fix all kinds of mysterious crashes that people have been encountering.  We have already sorted out a few crashing packages internally thanks to this.

Both of you are great - please exercise these features and help us make sure that they can be deployed well, and with minimal pain to the rest of everyone.

ocefpaf:
>  I will be pushing internally to make the error message more informative.

Thanks @msarahan!

> IMHO, features are well-worth adopting, because lining up the VC versions and keeping them consistent will fix all kinds of mysterious crashes that people have been encountering.

I agree.  However, for our particular case, since we build in a "controlled" environment (AppVeyor) we do not suffer (much) from such crashes.

PS: We are seeing some failures like [this one ](https://ci.appveyor.com/project/comtbot/conda-recipes/build/1.0.1292) now... :unamused:

pelson:
Great info @msarahan - thank you. Is there anywhere there is an issue on conda/conda? This kind of thing is actually a change in behaviour, since perviously Python would have just been downgraded right?

PR #554, opened 2015-11-02
Made the SciTools recipes point at the canonical SciTools/conda-recipes-scitools.

pelson:
Experimenting with removing the redundancy.

Another option was to use git submodules. The pros/cons:
 * pros
    * standard thing (doesn't need a special script)
    * could use symlinks into the correct location
    * readable on github
 * cons
    * needs updating everytime the conda-recipes-scitools repo is updated
    * awkward if we want to control specific commits for specific recipes (e.g. iris: master; cartopy: commit ab12cdef)
    * needs symlinks (these do work on git for windows though)


ocefpaf:
We tried submodules before and failed miserably.  The main issues were:

1. The added complexity to maintain the repo.  At the time we had people cloning/building/uploading from their local machines.
2. Different software versions.  We had to use a specific iris checkout for a few months because none of the released versions worked for our wokflow.

Issue 1 is no longer a problem, but 2 might surface again.

That being said... I like this approach better!  It makes it easier to react or solve for issue 2 when it happens.

PS: We call `udunits2` and not `udunits` (version) 2.  I guess that is due to historical reasons when most scientists installed both versions side-by-side to satisfy other software dependencies.

We do not package `udunits-1`, so I guess that renaming is not a big deal.  I will do that before merging this.

pelson:
> We do not package udunits-1, so I guess that renaming is not a big deal. I will do that before merging this.

I'm happy either way on this. Whichever works best for you guys!

ocefpaf:
Ping @rsignell-usgs here since he uses udunits for longer than me.

rsignell-usgs:
I'm not sure what the question here is, but I don't know of anybody using the  udunits v1 anymore.  Everyone is using udunits v2.   http://www.unidata.ucar.edu/software/udunits/#documentation

ocefpaf:
The question is: how should we name the package `udunits` or `udunits2`?

rsignell-usgs:
Let's ask Unidata: @lesserwhirls, which shall it be: `udunits` or `udunits2`?

lesserwhirls:
It's not clear from my point of view. The way the units are defined between
v1 and v2 changed quite a bit, and I assume the code did as well.

That said, the github repo is udunits-2, so I would say udunit2 is the way
to go:

https://github.com/Unidata/UDUNITS-2

Cheers!

pelson:
Thanks @lesserwhirls & @rsignell-usgs. Done in https://github.com/SciTools/conda-recipes-scitools/pull/124.

ocefpaf:
I see `pyyaml` being updated, `pyyaml:          3.11-py34_1   --> 3.11-py34_2`, yet AppVeyor is failing with:

```python
Traceback (most recent call last): 
  File "scripts/expand_source.py", line 12, in <module>
    import yaml
ImportError: No module named yaml
Command exited with code 1
```

I can debug this later though. The `scripts/expand_source.py` seems to be working as expected.

@rsignell-usgs I am :+1: with this PR ideas.  However, the IOOS channel is your baby and you should make the call to merge.  I don't think that the situation  2 (https://github.com/ioos/conda-recipes/pull/554#issuecomment-152992952) is likely yo happen again.

rsignell-usgs:
Hey, if @pelson  and @ocefpaf agree, who am I to disagree?
Seriously though, this is a good idea.    I understand @ocefpaf can commit on @scitools recipes, so we've got it covered without duplication -- outstanding!

PR #563, opened 2015-11-06
Added mocsy

### LINKED ISSUE #550: Package mocsy
https://github.com/jamesorr/mocsy

ocefpaf:
Closes #550 

Ping @emiliom for testing.

### DIFF
--- a/mocsy/bld.bat
+++ b/mocsy/bld.bat
@@ -1,0 +1,10 @@
+set CFG=%USERPROFILE%\pydistutils.cfg
+echo [config] > "%CFG%"
+echo compiler=mingw32 >> "%CFG%"
+echo [build] >> "%CFG%"
+echo compiler=mingw32 >> "%CFG%"
+echo [build_ext] >> "%CFG%"
+echo compiler=mingw32 >> "%CFG%"
+
+"%PYTHON%" setup.py install --single-version-externally-managed --record record.txt
+if errorlevel 1 exit 1
--- a/mocsy/build.sh
+++ b/mocsy/build.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+if [ `uname` == Darwin ]; then
+    export CFLAGS="-arch x86_64"
+    export FFLAGS="-static -ff2c -arch x86_64"
+    export LDFLAGS="-Wall -undefined dynamic_lookup -bundle -arch x86_64"
+    export LDFLAGS="-Wl,-search_paths_first -L$(pwd) $LDFLAGS"
+fi
+
+$PYTHON setup.py install --single-version-externally-managed --record record.txt
--- a/mocsy/meta.yaml
+++ b/mocsy/meta.yaml
@@ -1,0 +1,30 @@
+package:
+    name: mocsy
+    version: "2.0.1"
+
+source:
+    git_url: https://github.com/ocefpaf/mocsy.git
+    git_tag: python_package
+
+build:
+    number: 0
+    #skip: True  # [win]
+
+requirements:
+    build:
+        - python
+        - setuptools
+        - mingwpy  # [win]
+        - numpy
+    run:
+        - python
+        - numpy
+
+test:
+    imports:
+        - mocsy
+
+about:
+    home: https://github.com/jamesorr/mocsy.git
+    license: MIT License
+    summary: 'Routines to model ocean carbonate system thermodynamics'
--- a/scripts/run_docker_build.sh
+++ b/scripts/run_docker_build.sh
@@ -31,7 +31,7 @@ export CONDA_NPY='19'
 export PYTHONUNBUFFERED=1
 echo "$config" > ~/.condarc
 
-conda install obvious-ci -c conda-forge
+conda install --yes obvious-ci -c conda-forge
 
 # A lock sometimes occurs with incomplete builds.
 # The lock file is stored in build_artefacts.
@@ -41,7 +41,6 @@ python /conda-recipes/scripts/expand_source.py
 
 conda install --yes anaconda-client
 conda info
-unset LANG
 
 obvci_conda_build_dir /conda-recipes $UPLOAD_OWNER --build-condition "numpy >=1.9" "python >=2.7,<3|>=3.4,<3.5"
 


emiliom:
Thanks!! I don't have admin/write rights on this repo, so I can't merge. But can you say more about the status? Is the [pull request you submitted to the source mocsy repo](https://github.com/jamesorr/mocsy/pull/1) required before this conda recipe can work, or is this first version of a mocsy conda recipe ready to go with the mocsy code as-is?

If it's the latter (the conda recipe is ready as is), I can start testing soon after this pull request is merged and the conda package is ready on the ioos channel.

ocefpaf:
>  required before this conda recipe can work, or is this first version of a mocsy conda recipe ready to go with the mocsy code as-is?

The conda package points to the branch used in the PR https://github.com/jamesorr/mocsy/pull/1. Eventually we will point the conda package back to the original source.

I am still fixing some things here...  CircleCI is green but not OK.  (I cannot find out what is wrong with.)

But  I will merge it soon though.

PS: I am surprised I got AppVeyor working on Win64, but somehow it fails for python 3 :unamused:

emiliom:
> The conda package points to the branch used in the PR jamesorr/mocsy#1. 

Duh! Sorry, I missed this on meta.yaml: `git_url: https://github.com/ocefpaf/mocsy.git`

> But I will merge it soon though.

Thanks.

> PS: I am surprised I got AppVeyor working on Win64, but somehow it fails for python 3

Me too! FYI, python 3 support is not important at this initial stage, I think. Though it'll be nice to have.

ocefpaf:
> python 3 support

The funny thing is that there is nothing "python 3" in the code.  It seems like a bug in numpy distutils that does not know about mingwpy.

emiliom:
That's sad ...

ocefpaf:
@emiliom the binaries will show up in the channel soon.  Here is a (very) simple notebook that uses mocsy:

https://github.com/ocefpaf/mocsy/blob/python_package/notebooks/mocsy.ipynb

We need to wrap those Fortran function as python functions that:

- hide the private methods
- renamed them (removing the `m` at the beginning)
- take care of the shape and size of the input.

ocefpaf:
@emiliom quick question: is the EOS-80 the right way to go here?

emiliom:
Awesome!! I'll add working example notebooks based on the examples already available from the mocsy web site. I'll also follow up with James Orr next week, on the pull request you submitted.

Re: EOS-80 question. I don't know. Are you saying the code uses EOS-80, or that in your suggestions to "Re-factor some parts of the Fortran code as pure python and/or make use of pure python module currently available (like seawater)" EOS-80 seems like the logical choice? If it's the former, I wouldn't know what's best, except to say that mocsy and most other carbonate-system packages strive to implement the best practices recommended by Andrew Dickson (Scripps) and colleagues; it's possible that those best practices use EOS-80 b/c that was deemed to be the most appropriate one at the time, even if that's not the case now. But then switching to a more recent equation of seawater (TEOS-10?) would break that strict adherence to the Dickson best practices, which would break comparability and consistency.

ocefpaf:
> I'll also follow up with James Orr next week, on the pull request you submitted.

Thanks!

> Are you saying the code uses EOS-80

Yes. EOS-80 is present in virtually all CO2-code I see out there not only mocsy!  In mocsy you can see the original Fortran files like [this one](https://github.com/jamesorr/mocsy/blob/master/sw_adtg.f90) for example.

> it's possible that those best practices use EOS-80 

I suspected something like that because I do see EOS-80 everywhere.

>  But then switching to a more recent equation of seawater (TEOS-10?) would break that strict adherence to the Dickson best practices, which would break comparability and consistency.

That is why I asked.  In case we decided to re-factor we need to know if we stick with EOS80 or not.  (I am ignorant on this stuff, but I will ask Juliana once she gets home ;-)

emiliom:
Got it. I'm no expert on the CO2 algorithms, but I do know that adherence to Dickson's best practices is very important; it's the gold standard. So if EOS-80 is part of that best practice, we shouldn't mess with it initially. BTW, thanks for pointing me to one of the Fortran files that has EOS-80.

emiliom:
And [mocsy is now in the IOOS channel](https://anaconda.org/ioos/mocsy)! Woo-hoo! Thanks so much.

Unfortunately I have other things I need to work on today, so I probably won't get to try mocsy until tomorrow, or possibly Monday. I have a workshop in Portland all day Sunday.

emiliom:
I couldn't resist testing mocsy. I created a simple conda env with just mocsy and pandas (and other really basic things). Everything seems to work! I'm reproducing all the examples from http://ocmip5.ipsl.jussieu.fr/mocsy/pyth.html that don't require other dependencies. I'm able to run everything in Examples 1 & 2, and the values seem reasonable (I found a bug in how an argument was specified in Example 2, but it was an easy fix).

Next week I'll make those examples cleaner and easier to read, and will use pandas where currently a collection of numpy arrays is being used. When that's ready I'll share it with you, load the notebook somewhere (maybe a gist, or maybe I'll fork your mocsy repo, add this notebook, and send you a pull request). Then, once that notebook is ready, I'll contact James Orr.

This is awesome! FYI, I've also contacted the modeling group that I'm working with, to see if they're interested in using mocsy. They were planning to use CO2SYS-matlab, but will consider mocsy. They'll make a decision in about a week, so this is super timely. That group uses Python quite a bit.

ocefpaf:
> Next week I'll make those examples cleaner and easier to read, and will use pandas where currently a collection of numpy arrays is being used. When that's ready I'll share it with you, load the notebook somewhere (maybe a gist, or maybe I'll fork your mocsy repo, add this notebook, and send you a pull request). Then, once that notebook is ready, I'll contact James Orr.

Awesome!  I want to make unittests out of those examples too.  Maybe we can augment PR https://github.com/jamesorr/mocsy/pull/1 or do that in a new PR.

emiliom:
> Awesome! I want to make unittests out of those examples too. Maybe we can
augment PR jamesorr/mocsy#1 or do that in a new PR.

Nice. A separate PR would be better, I think.

emiliom:
I forked your mocsy repo and created two notebooks (on a new branch):
- [mocsy_docs.ipynb:](https://github.com/emiliom/mocsy/blob/python_package_em/notebooks/mocsy_docs.ipynb) like your own notebook, but with more documentation print outs and some additional introductory info
- [mocsy_examples.ipynb:](https://github.com/emiliom/mocsy/blob/python_package_em/notebooks/mocsy_examples.ipynb) Implements [Examples 1-3 from the mocsy Python web page](http://ocmip5.ipsl.jussieu.fr/mocsy/pyth.html). I found and fixed a couple of small but fatal mistakes in Examples 2 & 3, involving the input argument `patm`. In those two examples, I also restructured the mvars call to make it more readable. Finally, I added code to read the mvars output (plus the input variables) into a Pandas DataFrame and created a couple of plots. For Example 3, instead of reading a local file I'm now reading the file from the mocsy source repo.

Let me know what you think, and if you have any suggested changes.

Everything is good, with one big exception: pCO2 output seems to be corrupted at depths below the surface layer. It's strange, b/c most other variables (I didn't check all) are fine. This validation is possible for Example 3 b/c the repo included the output csv file corresponding to the input csv file.

emiliom:
A couple of additional comments:
- I assume that creating the single `_mocsy.f90` from the original files is a necessary step? This is not a problem as long as we can generate it automatically (we can write a script) from the original files when the mocsy fortran code is updated. I don't think it'd be a good idea to ask the package authors to restructure their code themselves just to make the Python wrapper happy.
- The doc strings that are returned are fairly generic and unhelpful, specially for the input and output arguments (no definition of the variables,units, options, etc). But the original f90 files do have very useful function doc strings. It'd be really helpful if those doc strings could be made fully accessible in the Python package. Can this be done?
- I haven't investigated where the pCO2 error may be coming from. I don't know if the best strategy is to have the mocsy authors confirm the output from their own Python wrapper (and Fortran code), or for us to track the issue ideally with their help. I don't read Fortran any more; that'd be fairly painful for me, from what I can see.

ocefpaf:
> Let me know what you think, and if you have any suggested changes.

These are awesome!  The notebooks should be part of examples/docs for the python module.  Looking at the notebooks I can see some obvious optimizations that we can do from the Python side, like:

- When wrapping the functions we should return a float and not an array for single numbers.
- The function calls have many `kwargs` that start with `op+<some letters>`.  We can be more verbose and  clear, thanks to auto-complete, and we can also remove the `op`.
- Add some defaults to those options would also help.

I am sure that, as we start using the module, more things will show up.  BTW these are issues that we should discuss in https://github.com/jamesorr/mocsy.

> pCO2 output seems to be corrupted at depths below the surface layer.

We need to confirm if this bug was introduced by the packaging process (possible but unlikely) or if it was already there.  (Another case for adding unittest ASAP!) 

> I assume that creating the single `_mocsy.f90` from the original files is a necessary step?

Not necessarily...  I was lazy there and went for the easiest route.  My point in doing that was to convince the original author to lay a plan out to get most of that code outside Fortran.  I believe that ~60% is already done (the EOS-80 stuff).  Then the Fortran part would be small enough to be a single file.

> The doc strings that are returned are fairly generic and unhelpful, specially for the input and output arguments (no definition of the variables,units, options, etc). But the original f90 files do have very useful function doc strings. It'd be really helpful if those doc strings could be made fully accessible in the Python package. Can this be done?

Those docstrings are generated automatically by the Fortran wrapping process. (Looking from that perspective they are pretty awesome ;-).

New docstring should be part of the Python wrapping plan I mentioned above, together with a better input/output parsing and `kwargs`.  And when doing that we should copy from the Fortran comments.

> I haven't investigated where the pCO2 error may be coming from. I don't know if the best strategy is to have the mocsy authors confirm the output from their own Python wrapper (and Fortran code), or for us to track the issue ideally with their help. I don't read Fortran any more; that'd be fairly painful for me, from what I can see.

We should open an issue just for this and see what they have to say.

emiliom:
Thanks! Great feedback. I'll contact James Orr late today or tomorrow at the latest. I'll first bring up our conda packaging and the pCO2 error. I see those two as the top priorities. FYI, I may go via email first (cc'ing you); I have a hunch that email will get us a faster response compared to github issues or pull requests.

Yup, the wrapper could definitely use some Python love, including the ones you mentioned. I won't put any effort into that until the pCO2 error is fixed and we have buy in from James.

ocefpaf:
Sounds like a plan!

jamesorr:
The efforts of both of you regarding mocsy are much appreciated!  Concerning some of the issues you have raised, here are a few comments:

0) The notebooks are excellent!  Much appreciated.

1) Yes, mocsy uses EOS-80 and not TEOS-10 because it was the former that was in vigor when the best practices were defined (Dickson et al., 2007).  Furthermore, I would not change everything to TEOS10, not yet anyway, because many ocean carbon cycle models still use EOS80. A better option may be to provide functions to convert between TEOS-10 and EOS-80, but this may already be found elsewhere.

2) The large subsurface pCO2 differences you are seeing is not an error.  See the documentation of the OPTGAS function in the fortran code.  Results for pCO2 vary greatly as a function of that option, as detailed in the final publication describing mocsy v2.0 (Fig. 1, section 2.1.3):
http://www.geosci-model-dev.net/8/485/2015/gmd-8-485-2015.html

3) The docstrings in mocsy are those extracted automatically by f2py from the original fortran code.  Perhaps they can be improved with options to f2py; I'll need to check.

Finally, note that 'mocsy' is first and foremost designed for ocean models making carbonate chemistry calculations ONLINE, i.e., during model runs (in Fortran). The OFFLINE calculations, which mocsy allows to be done conveniently in python are a big plus, but I would really like to avoid calling routines in python (e.g., from 'seawater') because that would break the compatibility between the Fortran and python use of mocsy.

ocefpaf:
@jamesorr  and @emiliom let's take this thread to https://github.com/jamesorr/mocsy/issues?

PR #564, opened 2015-11-06
Add jinja2 to compliance-checker requirements

daf:
Became a dep upstream on Jul 10th 2015

### DIFF
--- a/compliance-checker/meta.yaml
+++ b/compliance-checker/meta.yaml
@@ -23,6 +23,7 @@ requirements:
         - cf_units
         - requests
         - python-dateutil
+        - jinja2>=2.7.3
     run:
         - python
         - owslib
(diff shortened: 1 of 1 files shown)


compliance-checker/meta.yaml
```diff
@@ -23,6 +23,7 @@ requirements:
         - cf_units
         - requests
         - python-dateutil
+        - jinja2>=2.7.3
```
ocefpaf: Need a space after `jinja2`

compliance-checker/meta.yaml
```diff
@@ -31,6 +32,7 @@ requirements:
...
```
ocefpaf: Same as above.

ocefpaf:
Thanks @daf.  You also need to bump the build number.

PR #575, opened 2015-11-09
Phreeqpy

### LINKED ISSUE #551: Package phreeqpy
http://www.phreeqpy.com/

ocefpaf:
The PyPI tarball ships the c-libs :scream: :scream_cat: :scream: :scream_cat: 

Closes #551

Ping @emiliom

### DIFF
--- a/iphreeqc/build.sh
+++ b/iphreeqc/build.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+./configure --prefix=$PREFIX
+
+make
+make check
+make install
--- a/iphreeqc/meta.yaml
+++ b/iphreeqc/meta.yaml
@@ -1,0 +1,24 @@
+package:
+    name: iphreeqc
+    version: 2.18.4
+
+source:
+    fn: iphreeqc-2.18.4-6386.tar.gz
+    url: ftp://brrftp.cr.usgs.gov/pub/charlton/iphreeqc/iphreeqc-2.18.4-6386.tar.gz
+    md5: cfa673561cc8851da2ef3de9414e3318
+
+build:
+    number: 0
+    skip: True  # [win]
+
+requirements:
+    build:
+    run:
+
+#test:
+    #commands:
+
+about:
+    home: http://www.phreeqpy.com/
+    license: GNU General Public License v2
+    summary: "A Computer Program for Speciation, Batch-Reaction, One-Dimensional Transport, and Inverse Geochemical Calculations"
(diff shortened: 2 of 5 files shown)


emiliom:
Wow! I'm not sure if all that :scream: means that "The PyPI tarball ships the c-libs" is a good or bad thing ... Based on the [phreeqpy installation instructions](http://www.phreeqpy.com/), it sounds like a good thing.

Before you sink a lot of time into this, it's worth reminding you that:
- I think this is awesome.
- But I don't have plans to use phreeqpy in the near future. I'm willing to [create and run a simple test](https://gist.github.com/emiliom/7866654) to check that the conda package works, but that's about it.
- PHREEQC is an important and extensively used geochemical library, but probably not in oceanography, I think. It's [actively maintained.](http://wwwbrr.cr.usgs.gov/projects/GWC_coupled/phreeqc/status.htm)
- As I mentioned before, the phreeqpy wrapper seems abandoned. But I'm willing to contact the author to ask about its status.

ocefpaf:
> Based on the phreeqpy installation instructions, it sounds like a good thing.

It depends on your point of view.  Some people would never accept a pre-compiled library to be shipped in a source distro.  (And the actually library was not hard to compile.  I have `iphreeq` on this same PR.)

> As I mentioned before, the phreeqpy wrapper seems abandoned. But I'm willing to contact the author to ask about its status.

Don't worry. I don't have plans to maintain this recipe.  I only added the package because it was doable and I know at least 2 people that will try this.

emiliom:
> Don't worry. I don't have plans to maintain this recipe. I only added the package because it was doable and I know at least 2 people that will try this.

Ok, I'm good with that.

FYI, PHREEQC (not phreeqpy specifically) seems to be under continued, regular development. They just published a new article on one of its linked modules (["PhreeqcRM: A reaction module for transport simulators based on the geochemical model PHREEQC", Sept. 2015](http://dx.doi.org/10.1016/j.advwatres.2015.06.001)), and that article is currently the [8th most downloaded article in Advances in Water Resources, where it was published](http://www.journals.elsevier.com/advances-in-water-resources/most-downloaded-articles/).

ocefpaf:
> FYI, PHREEQC (not phreeqpy specifically) seems to be under continued, regular development.

That right there is another reason why bundling the c-libs is a bad choice.  People must know which version they are linking against.  `phreeqpy` bundles a really old version of (` 2.18.4` if I am not mistaken.)

emiliom:
Great point.

PR #590, opened 2015-11-13
Add recipes for OpenClimateGIS

bekozi:
This PR adds recipes for OpenClimateGIS and dependencies not covered by IOOS.

 * Building the "next" branch of OpenClimateGIS is recommended for netCDF4 and OSX compatibility.
 * Python 2.7 only.
 * The default conda Fiona package conflicts with GDAL 2.0. The Fiona release supporting 2.0 is in prep. In the meantime, the trunk is built.
 * OpenClimateGIS uses "cfunits" v. "cf_units"...
 * Don't have many ``.bat`` files.
 
The build script looks something like below. There are nested environment variables in the ``meta.yaml`` files. Is it better for your system to use embedded versions/tags?

```sh
export CBUILD_OCGIS_TAG=next
export CBUILD_OCGIS_VERSION=1.2.0.dev1
export CBUILD_FIONA_TAG=master
export CBUILD_FIONA_VERSION=1.6.2.dev1
export CBUILD_CFUNITS_TAG=v1.1
export CBUILD_CFUNITS_VERSION=1.1

conda build ocgis
# Optional dependency.
conda build cfunits
# Install with optional dependencies (no ESMF or ICCLIM yet).
conda install ocgis cfunits rtree
```

I tested building on a "continuum/anaconda" Docker image. I noticed you are using a CentOS6 image for your builds. We have used CentOS6 as a build image before without issue. Anyway, let's see how it goes!


ocefpaf:
@bekozi Thanks for the PR!

I will review it shortly.

ocefpaf:
> OpenClimateGIS uses "cfunits" v. "cf_units"...

Any chance I can convince you to switch?  The libraries should be pretty similar, but cf_units has a better community support.

ocefpaf:
> Don't have many .bat files.

Please add the syntax:

```yaml
build:
    skip: True  # [win]
```

You may also skip Python 3 by adding `# [win and py3k]`

cfunits/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+$PYTHON setup.py install --record record.txt
+$PYTHON test/run_tests.py
```
ocefpaf: Do you think you can add this test to the recipe directory?  That way conda runs it automatically without this call.

cfunits/meta.yaml
```diff
@@ -0,0 +1,30 @@
+package:
+  name: cfunits
+  version: {{environ['CBUILD_CFUNITS_VERSION']}}
```
ocefpaf: Our style is to avoid environment variable to compose the `yaml` file to increase the readability.  We want our users to browse the recipes and know what gets build.

How hard would it be to change these?

cfunits/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+source:
+  git_url: https://bekozi@bitbucket.org/cfpython/cfunits-python.git
```
ocefpaf: I am an ignorant on bitbucket, but is that your fork or the canonical source for cfunits-python?

We also try to be consistent with the original Software name. In this case seems to be `cfunits-python`, correct?

cfunits/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+  git_tag: {{environ['CBUILD_CFUNITS_TAG']}}
+
+build:
+  number: 1
```
ocefpaf: Please set the first builds to `0`.

cfunits/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+requirements:
+  build:
+    - python
+    - setuptools
```
ocefpaf: I see you already added `--record record.txt`.  Thanks!  All modules that use `setuptools` we build with:

`--single-version-externally-managed --record record.txt`

That makes our life much easier in the long run.

click-plugins/meta.yaml
```diff
@@ -0,0 +1,24 @@
+package:
+  name: click-plugins
+  version: 1.0.2
+
+source:
+  git_url: https://github.com/click-contrib/click-plugins.git
+  git_tag: 1.0.2
+
+build:
+  number: 1
```
ocefpaf: Please set to `0`

fiona/meta.yaml
```diff
@@ -0,0 +1,44 @@
+package:
+  name: fiona
+  version: {{environ['CBUILD_FIONA_VERSION']}}
+
+source:
+  git_tag: {{environ['CBUILD_FIONA_TAG']}}
+  git_url: https://github.com/Toblerity/Fiona
```
ocefpaf: If I got you you said you are using `master` because of `gdal 2.0.0`, right?

If so consider pinning `gdal to <2.0.0` in `ocgis` instead.  We avoid as much as we can to use `master/dev` versions.

fiona/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+build:
+  number: 1
```
ocefpaf: Again please set to `0`

fiona/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+requirements:
+  build:
+    - python
+    - setuptools
```
ocefpaf: When adding `--single-version-externally-managed` it is usually unnecessary to list run dependencies in the build section.  That said, if the package imports itself to get a version or something, then you need to list them...

libpq/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [ "$(uname)" == "Darwin" ]; then
+    export DYLD_LIBRARY_PATH=${PREFIX}/lib:${PREFIX}/lib64
+else
+    export LD_LIBRARY_PATH=${PREFIX}/lib:${PREFIX}/lib64
```
ocefpaf: We should ask about this in the conda mailing list.  Not sure how that will play out in the log run.

libpq/meta.yaml
```diff
@@ -0,0 +1,17 @@
+package:
+  name: libpq
+  version: 9.4.5
+
+build:
+  number: 1
```
ocefpaf: Again `0` :wink:

libpq/meta.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+source:
+  fn: postgresql-9.4.5.tar.bz2
```
ocefpaf: Why do you call `postgresql` libpq?  Will the default channel version suffice?

http://anaconda.org/anaconda/postgresql

munch/meta.yaml
```diff
@@ -0,0 +1,26 @@
+package:
+  name: munch
+  version: 2.0.4
+
+source:
+  git_url: https://github.com/Infinidat/munch.git
+  git_tag: 2.0.4
+
+build:
+  number: 1
```
ocefpaf: :smile_cat:

ocgis/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+    number: 1
+
+requirements:
+    build:
+        - python
+        - setuptools
+        - numpy
+        - netcdf4
+        - shapely
+        - fiona =={{environ['CBUILD_FIONA_VERSION']}}
```
ocefpaf: If we can let's avoid the environment variables and the repetition of the run dependencies on the build section.

ocgis/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+    run:
+        - python
+        - numpy
+        - netcdf4
+        - shapely
+        - fiona =={{environ['CBUILD_FIONA_VERSION']}}
```
ocefpaf: You might want to pin `gdal` version here.

rsignell-usgs:
@bekozi, do you feel behind the :8ball: ?  :smile_cat:

fiona/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  run:
+    - python
+    - click
+    - six
+    - numpy
+    - gdal
+    - cligj
+    - munch
+    - click-plugins
+    - libpq # [osx]
```
ocefpaf: Should `cfunits` be here?

ocefpaf:
```sh
export CBUILD_OCGIS_TAG=next
export CBUILD_OCGIS_VERSION=1.2.0.dev1
export CBUILD_FIONA_TAG=master
export CBUILD_FIONA_VERSION=1.6.2.dev1
export CBUILD_CFUNITS_TAG=v1.1
export CBUILD_CFUNITS_VERSION=1.1
```

@bekozi We use Travis-CI, CircleCI, and AppVeyor to automatically build the recipes for OSX, Linux, and Windows respectively.  Everything is tied together with `obvious-ci`. Note that is `obvious-ci` who create the build matrix and takes care of the dependencies build order.

That said, it is better to avoid the environment variable and put them inside the `meta.yaml` files.

bekozi:
@rsignell-usgs: Ha! No one said it would be easy. :smiley:

@ocefpaf: I addressed the comments. Anything in question or requiring input is addressed below. Thanks for the thorough review!

> https://github.com/ioos/conda-recipes/pull/590#issuecomment-156512309
> Any chance I can convince you to switch? The libraries should be pretty similar, but cf_units has a better community support.

Yeah probably. I'll look into it. I'm guessing it's okay to keep pressing forward for now. You saw the ticket.

> https://github.com/ioos/conda-recipes/pull/590#discussion_r44819903
> Why do you call postgresql libpq? Will the default channel version suffice?

An oversight. That is all. :smiley:

> https://github.com/ioos/conda-recipes/pull/590#discussion_r44818668
> If so consider pinning gdal to <2.0.0 in ocgis instead. We avoid as much as we can to use master/dev versions.

We can go this route. There are complications with osx and GDAL versions < 2.0. See https://github.com/ContinuumIO/anaconda-issues/issues/380#issuecomment-124154635 for reference. For Linux there are no issues. I understand the desire to avoid development versions. Maybe we can get this working, and then I can publish a minor release we can build against (before merging)?

I am not sure when the ``fiona`` version supporting 2.0 will be out. They are working on it! That dev can then go away too.

Are you interested in supporting a GDAL build?

> https://github.com/ioos/conda-recipes/pull/590#discussion_r44820127
> You might want to pin gdal version here.

That should be taken care of in the ``fiona`` recipe?

> https://github.com/ioos/conda-recipes/pull/590#discussion_r44822629
> Should cfunits be here?

It's technically an optional dependency. However, I don't really see why ``rtree`` and ``cfunits-python`` should not be added as the impact is small. I had them on the original recipe.

I'll keep an eye on the CI builds too...

rsignell-usgs:
@bekozi , thanks for plugging away at this!  We'll get there!!!

bekozi:
The ``cfunits-python`` build is failing for Python 3, yet building Python 3 is supposedly disabled. What am I missing?

``py3k`` skipped: https://github.com/NESII/conda-recipes-ioos/commit/921087ebd5217b133be4265622cf74d6833406bd#diff-04a77a7db3f9089e7c58082e33011a0cR11

CI Failure: 

```sh
==================================== ERRORS ====================================
_____________________ ERROR collecting test/test_Units.py ______________________
../../envs/_test/lib/python3.5/site-packages/_pytest/python.py:584: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
../../envs/_test/lib/python3.5/site-packages/py/_path/local.py:650: in pyimport
    __import__(modname)
E     File "/opt/conda/conda-bld/work/test/test_Units.py", line 138
E       print 'cfunits-python version:', cf.__version__
E                                     ^
E   SyntaxError: Missing parentheses in call to 'print'
=========================== 1 error in 0.10 seconds ============================

TESTS FAILED: cfunits-python-1.1-py35_0 ./scripts/run_docker_build.sh returned exit code 1
```

ocefpaf:
> The cfunits-python build is failing for Python 3, yet building Python 3 is supposedly disabled. What am I missing?

I forgot to mention that you have to add `python >=2.7,<3` in the build section.

cfunits-python/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+  git_tag: v1.1
+
+build:
+  number: 0
+  skip: True # [win and py3k]
+
+requirements:
+  build:
+    - python
+    - udunits2
```
ocefpaf: Question: `cfunits-python` does not use setuptools?

cfunits-python/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+  name: cfunits-python
+  version: 1.1
+
+source:
+  git_url: https://bitbucket.org/cfpython/cfunits-python.git
+  git_tag: v1.1
+
+build:
+  number: 0
+  skip: True # [win and py3k]
```
ocefpaf: A case for `cf_units` :wink: 
It does work on Python 3.

ocefpaf:
@bekozi we are almost there.  The OSX failure ([Travis-CI](https://travis-ci.org/ioos/conda-recipes/builds/91051427)) is because cfunits is failing to find the udunits2 shared library.

Finding shared libraries in conda and load them with ctypes is a known issue.  The solutions are to use a `setup.cfg`, like we do for [cf_units](https://github.com/SciTools/conda-recipes-scitools/blob/master/cf_units/build.sh#L3-L12), or to patch the Software to forcing it to point to the conda version, like we do for shapely, [rtree](https://github.com/ioos/conda-recipes/blob/master/rtree/find_libray_nix.patch), and many others here.

cfunits-python/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+$PYTHON setup.py install --record record.txt
```
ocefpaf: Please add the ` --single-version-externally-managed` too.

fiona/meta.yaml
```diff
@@ -0,0 +1,42 @@
+package:
+  name: fiona
+  version: 1.6.2.dev1
+
+source:
+  git_tag: master
```
ocefpaf: If you are using master only because of `gdal 2.0.0`>  If so let's pin gdal to `< 2.0.0`.  We must avoid as much as we can using master/dev releases.  that brings all sort of instabilities and crashes for other packages that depends on it.

ocgis/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+        - cfunits-python
+        - rtree
+
+test:
+  requires:
+    - nose
+
+about:
+    home: http://ncpp.github.io/ocgis/
+    license: NCSA License
```
ocefpaf: Is this an [OSI approved](http://opensource.org/licenses) license?

cfunits-python/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+requirements:
+  build:
+    - python >=2.7,<3
+    - udunits2
+    - netcdf4
+  run:
+    - python >=2.7,<3
```
ocefpaf: Use that restriction only on `build` as I mentioned in my previous comment.

The reason for this is because conda has some weird behavior and removes The Python build string if you specify on both `build` and `run`.

fiona/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  build:
+    - python
+    - six
+    - numpy
+    - gdal
+    - cligj
+    - cython
+    - munch
+    - click-plugins
+    - postgresql # [osx]
```
ocefpaf: I am curios why this dependency is only needed on OSX.

cfunits-python/build.sh
bekozi: ``cfunits-python`` does not use ``setuptools``. I sent them a PR for it some time ago, but they want to stick with ``distutils``.

ocefpaf: No problem.  But py3k, setuptools, and easy udunits support via configparse seems like good reasons to switch to `cf_units` :wink:

ocgis/meta.yaml
bekozi: http://opensource.org/licenses/NCSA

cfunits-python/build.sh
rsignell-usgs: +1 for switching to `cf_units`.  We went through a lot of pain with the others, and `cf_units` has the brightest future, for sure, now that the British Met Office is on board.

bekozi:
Linux builds are passing (finally). There is one failure on ``appveyor``, but that is related to ``pyproj``. I'm not sure if you want to merge this, or wait until I have tagged an ``ocgis`` release that will better support OSX (i.e. ``cf_units``). The next release will also relax the ``netcdf4`` version constraint and, depending on the timing, allow for GDAL 2.0. Let me know!

Regardless, thanks for the assistance and patience. This is a great build system!

ocefpaf:
> Linux builds are passing (finally)

:tada: 

> There is one failure on appveyor

Never mind AppVeyor.  You are not building anything on Windows.  I already sent a fix [upstream](https://github.com/jswhit/pyproj/pull/32).  Pyproj is built against master :scream: and the failure is due to a new PR that introduced a bug.  (See... That is why we don't build against master :wink:)

> I'm not sure if you want to merge this, or wait until I have tagged an ocgis release that will better support OSX 

We can merge as is and people can start using `ocgis` on Linux.  We just need to open an issue for the OSX support to keep track of what needs to be done.

> Regardless, thanks for the assistance and patience.

Thank you for the perseverance!

> This is a great build system!

Indeed.  The credits shouldgo to @pelson who started all this for the [scitools](https://github.com/SciTools/conda-recipes-scitools).

Looking forward to get the recipes for the ESMF working too.

bekozi:
Awesome. Thanks!

ocgis/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    skip: True # [py3k]
+
+requirements:
+    build:
+        - python >=2.7,<3
+        - cfunits-python
+        - fiona
+        - gdal <2
+        - libgdal <2
+        - netcdf4 <=1.1.6
```
ocefpaf: @bekozi I forgot.  What is the reason behind a `netCDF4-python <= 1.1.6`?

bekozi: There is an odd issue with netCDF4-python v1.1.6+ that required this (very specific) workaround: https://github.com/NCPP/ocgis/blob/176a86e293e101877de1098a35f9efde73e83718/src/ocgis/interface/base/attributes.py#L32. I had trouble isolating the issue, but it seems like there is a rogue pointer somewhere in the netCDF4 attribute code. The workaround came after the 1.2 release, and it is fixed in the upcoming release. If this is causing a problem, I could add a patch file for the interim?

ocefpaf: We can wait for a new release.  I was just wondering why.  Thanks for the explanation.

PS: This surfaced because I am re-building `gdal 1.11.3` and `fiona` and I noticed that, enforcing `netcdf4 <= 1.1.6` here, we end up with `numpy < 1.10`.  That happens because there is not netCDF4-python 1.1.6 compiled against a newer numpy.

bekozi: Got it. Release should be next month. An IOOS-hosted ``gdal`` and ``fiona`` will be awesome!

PR #597, opened 2015-11-17
Adding launcher

rsignell-usgs:
Adding `launcher` so that folks can use it from the ioos environment


ocefpaf:
We don't need to wait for the CIs on this one.

PR #599, opened 2015-11-18
Add Linux recipes for ESMF and ESMPy

bekozi:
I tested the recipes using the ``continuumio/anaconda`` Docker image:

```sh
apt-get update
apt-get install git build-essential gfortran
conda build esmpy
```

Builds and ``ESMPy`` tests worked fine. I did not test on an external container. I went for very slim recipes. Let's see how it goes!

Of note: 
 * ``ESMPy v7`` will be coming out soonish causing changes to the testing. We can test externally for example.
 * It looks like conda folks are considering adding a ``--depth`` flag to ``git clone`` which would reduce the size of the ``ESMF`` repo: https://github.com/conda/conda/issues/1262.

Things to do before merge:
 - [ ] Enable ``ESMF`` testing. I think it makes sense to get the basic build and installation processes working. The tests take 20-30 minutes to run. ``ESMPy`` tests will run in a much shorter time. If ``ESMF`` tests are needed in addition to ``ESMPy`` tests is open for debate.
 
Watchers: @cdeluca, @rokuingh

### DIFF
--- a/esmf/build.sh
+++ b/esmf/build.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+
+export ESMF_DIR=`pwd`
+export ESMF_INSTALL_PREFIX=${PREFIX}
+export ESMF_NETCDF="split"
+export ESMF_NETCDF_INCLUDE=${PREFIX}/include
+export ESMF_NETCDF_LIBPATH=${PREFIX}/lib
+export ESMF_COMM=mpich2
+
+make  -j ${CPU_COUNT}
+# TODO (bekozi): Enable all ESMF tests.
+#make all_tests | tee ~/esmf_all_tests.out
+make install
--- a/esmf/build.sh
+++ b/esmf/build.sh
@@ -9,6 +9,6 @@ export ESMF_NETCDF_LIBPATH=${PREFIX}/lib
 export ESMF_COMM=mpich2
 
 make  -j ${CPU_COUNT}
-# TODO (bekozi): Enable all ESMF tests.
-#make all_tests | tee ~/esmf_all_tests.out
+# TODO (bekozi): Enable ESMF tests?
+#make all_tests
 make install
--- a/esmf/meta.yaml
+++ b/esmf/meta.yaml
@@ -6,7 +6,6 @@ build:
   number: 0
   skip: True # [osx]
   skip: True # [win]
-  skip: True # [py3k]
 
 source:
   git_url: git://git.code.sf.net/p/esmf/esmf
--- a/esmf/meta.yaml
+++ b/esmf/meta.yaml
@@ -4,9 +4,6 @@ package:
 
 build: 
   number: 0
-  rpaths:
-    - lib/
-    - lib64/
   skip: True # [osx]
   skip: True # [win]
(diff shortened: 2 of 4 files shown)


esmf/build.sh
```diff
@@ -0,0 +1,14 @@
...
+export ESMF_DIR=`pwd`
+export ESMF_INSTALL_PREFIX=${PREFIX}
+export ESMF_NETCDF="split"
+export ESMF_NETCDF_INCLUDE=${PREFIX}/include
+export ESMF_NETCDF_LIBPATH=${PREFIX}/lib
+export ESMF_COMM=mpich2
+
+make  -j ${CPU_COUNT}
+# TODO (bekozi): Enable all ESMF tests.
+#make all_tests | tee ~/esmf_all_tests.out
```
ocefpaf: No need to output to a file here.  This will happen inside the CIs.

esmf/meta.yaml
```diff
@@ -0,0 +1,29 @@
+package:
+  name: esmf
+  version: 6.3.0rp1
+
+build: 
+  number: 0
+  rpaths:
+    - lib/
+    - lib64/
```
ocefpaf: Do you need both?

ocefpaf:
For some reason CircleCI is skipping `esmf` and trying to build `esmpy` without it.

ocefpaf:
@bekozi thanks for the PR.  I will take a better look at what is going on later today or early tomorrow.  I am a little bit swamped here.  (Moving to a new place.)

bekozi:
@ocefpaf good luck with the move!

I removed the ``rpaths`` from the ``ESMF`` meta file. My untrained eye could not determine what is causing the CircleCI conflict...

esmf/meta.yaml
```diff
@@ -0,0 +1,26 @@
+package:
+  name: esmf
+  version: 6.3.0rp1
+
+build: 
+  number: 0
+  skip: True # [osx]
+  skip: True # [win]
+  skip: True # [py3k]
```
pelson: This line could be the cause of the skip. esmf doesn't depend on Python, so no need for this skip.

bekozi:
@pelson good catch. The builds are working now.

ocefpaf:
Thanks @bekozi and @pelson.

Phil I believe you are interested in ESMF too, right?  Should be send these to conda-forge or scitools?

pelson:
> Phil I believe you are interested in ESMF too, right? 

Yes, very much so.

> Should we send these to conda-forge or scitools?

The fact that they are *available* is good for me at this stage. The obvious place will be conda-forge once we have the feedstock concept up and running (I'm getting there slowly :wink:).

Thanks for submitting these recipes :+1:

ocefpaf:
>> Phil I believe you are interested in ESMF too, right?
>
> Yes, very much so.
>>
>> Should we send these to conda-forge or scitools?
>
> The fact that they are available is good for me at this stage. The
obvious place will be conda-forge once we have the feedstock concept up and
running (I'm getting there slowly ).

I've been following those changes, but did not try them yet. I will be out
a week or so. Once I get back I will send the feedstocks.

PR #600, opened 2015-11-19
Adding r-oce (and r-gsw)

ocefpaf:
These are R packages useful for teaching Python as a glue language.


ocefpaf:
These will probably fail because they need the `conda r channel`.  I am not sure if it is worth adding those here yet.  This recipe is just a place holder.

PR #602, opened 2015-11-19
Add RISE

ocefpaf:
Jupyter/IPython notebooks slide extension


ocefpaf:
RISE needs to access the local config and conda points it to `/root/.jupyter/nbconfig`, so this is a no no recipe :unamused:

PR #606, opened 2015-11-20
Add luigi 2.0.0 and python-daemon 2.0.6

kwilcox:
These are OSX/Linux only

### DIFF
--- a/luigi/meta.yaml
+++ b/luigi/meta.yaml
@@ -9,6 +9,7 @@ source:
 
 build:
     number: 0
+    skip: True  # [win]
     entry_points:
         - luigi = luigi.cmdline:luigi_run
         - luigid = luigi.cmdline:luigid
--- a/luigi/meta.yaml
+++ b/luigi/meta.yaml
@@ -1,0 +1,46 @@
+package:
+    name: luigi
+    version: "2.0.0"
+
+source:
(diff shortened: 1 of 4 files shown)


ocefpaf:
Kyle I will be offline for a while and I cannot merge this using my
not-so-smart the phone.

Feel free to merge it or ping Rich.

kwilcox:
How do I go about skipping windows builds?  I left out `bld.bat` in hopes that would do the trick but AppVeyor is still trying to import.  Should I add the `nowin` comment to the import and test sections of `meta.yml`?

ocefpaf:
Take a look at the nco meta.yaml. You need to add the "skip build" part.

luigi/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+    name: luigi
+    version: "2.0.0"
+
+source:
+    fn: luigi-2.0.0.tar.gz
+    url: https://pypi.python.org/packages/source/l/luigi/luigi-2.0.0.tar.gz
+    md5: 06258afcfcdd2f829167450fd5fed604
+
+build:
+    number: 0
```
rsignell-usgs: add 
   skip: True  # [win]
here as in:
https://github.com/ioos/conda-recipes/blob/master/nco/meta.yaml#L9-L11

PR #621, opened 2015-11-30
Minor tweaks

ocefpaf:
No need to bump build number.  Just minor tweaks to the recipe.


kwilcox:
I'll start indenting with 4 spaces and clearing the build dependencies going forward (I generated this one with `conda template`)!

ocefpaf:
No problem.  It is faster to just merge and change this later rather than reviewing and waiting for the CIs.

Note that the binaries are the same.  These changes are just to keep a consistent style of the recipes.

(We indent using 4 spaces and try to keep the build dependencies as minimal as possible.)

PR #626, opened 2015-12-03
Test netCDF4 build on Windows

ocefpaf:
~~Just testing.~~

The default channel is outdated (version 1.1.9 is two releases behind) and not build with Cython (so no `filepath` method).

### DIFF
--- a/netcdf4/bld.bat
+++ b/netcdf4/bld.bat
@@ -1,0 +1,12 @@
+set SITECFG=%SRC_DIR%/setup.cfg
+
+echo [options] > %SITECFG%
+echo use_cython=True >> %SITECFG%
+echo [directories] >> %SITECFG%
+echo HDF5_libdir = %LIBRARY_LIB% >> %SITECFG%
+echo HDF5_incdir = %LIBRARY_INC% >> %SITECFG%
+echo netCDF4_libdir = %LIBRARY_LIB% >> %SITECFG%
+echo netCDF4_incdir = %LIBRARY_INC% >> %SITECFG%
+
+"%PYTHON%" setup.py install --single-version-externally-managed  --record record.txt
+if errorlevel 1 exit 1
--- a/netcdf4/meta.yaml
+++ b/netcdf4/meta.yaml
@@ -8,7 +8,6 @@ source:
     md5: 9d9a7015ee98ec6766adc811d95b82c3
 
 build:
-    skip: True  # [win]
     number: 1
     entry_points:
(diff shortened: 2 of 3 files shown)


ocefpaf:
Ping @rsignell-usgs can you test this on Windows?

@kwilcox I am certain this is not a blocker for you, but `wera2netcdf` (or any other package that depends on netCDF4) is not available on Windows+Python 3.5

pelson:
Very nice. :+1:

ocefpaf:
I guess this should be in conda-forge.

It is sad to see some key packages lagging behind in the default channel.  We really need to strengthen the community channels and find a way to push these key packages back into the default channel.

pelson:
> We really need to strengthen the community channels and find a way to push these key packages back into the default channel.

Maybe one day the community channel(s) should be part of the default channel. We need to get our house in order first though - I'm currently doing a *lot* of conda stuff at the moment, but none of it is directly targeted at the community channel (aka conda-forge) stuff. Once I've got my current project out of the way (which is funding some of the tools which will make conda-forge better/easier) I'll start spinning the community channel work up again in earnest. I don't expect it will be quick, but the groundswell has started and I think we will be in a very strong position in 3-6 months.

rsignell-usgs:
@ocefpaf , how would you like this tested?

ocefpaf:
See if you can use `nc.filepath()` and the new stuff like the `get_variables_by_attributes`.

ocefpaf:
> Maybe one day the community channel(s) should be part of the default channel. 

I would love for the community channel(s) to become a "vetting step" to add packages to the default channel.

>  I'm currently doing a *lot* of conda stuff at the moment, but none of it is directly targeted at the community channel (aka conda-forge) stuff.

I am following from a safe distance.  You last project, conda-execute, made me spend 2 days re-writing my scripts :stuck_out_tongue_winking_eye:  (Totally worth it BTW!)

> I think we will be in a very strong position in 3-6 months.

I think that is quick!

rsignell-usgs:
@ocefpaf , reporting back that the new netcdf4 functionality `nc.filepath()` and  `get_variables_by_attributes` work fine on Windows using python 2.7 and 3.4.   

See cell [20] and [21] here:
https://github.com/rsignell-usgs/notebook/blob/master/UGRID/FVCOM_depth_and_velocity.ipynb

:tada:

ocefpaf:
Awesome!  Thanks.

As soon as `mingwpy` sorts python 3.5 out I will build it. For now we are limited to python 3.4 on Windows.

PR #624, opened 2015-12-01
Fiona

ocefpaf:
Fiona is broken in the default channel.  It is compiled against  `gdal 2.0.0`, which is known to not work properly with fiona, and is missing two core dependencies: `munch` and `click-plugins`.

This PR re-introduces fiona in the ioos channel to avoid headaches like #623.

Issue #623, opened 2015-12-01
Fiona fails to open shapefiles (Geopandas/Fiona/GDAL)


fmaussion:
I hope this is the right place to describe the problem.

All my attempts to make Geopandas work correctly with conda failed on Linux, Windows and on Mac OSX.

```
conda create -n conda_test python=2.7
source activate conda_test
python 
```

```python
>>> import geopandas
>>> geopandas.read_file('/home/mowglie/disk/Data/GIS/SHAPES/world_borders/world_borders.shp')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mowglie/.bin/conda/envs/oggm_conda_test/lib/python2.7/site-packages/geopandas/io/file.py", line 20, in read_file
    gdf = GeoDataFrame.from_features(f, crs=crs)
  File "/home/mowglie/.bin/conda/envs/oggm_conda_test/lib/python2.7/site-packages/geopandas/geodataframe.py", line 173, in from_features
    for f in features:
  File "fiona/ogrext.pyx", line 1080, in fiona.ogrext.Iterator.__next__ (fiona/ogrext.c:17244)
  File "fiona/ogrext.pyx", line 161, in fiona.ogrext.FeatureBuilder.build (fiona/ogrext.c:3254)
IndexError: list index out of range
```
The problem is unrelated to the shapefile and/or the python version. I wonder if the problem doesn't occur at the GDAL level already.

Thanks for any tip.

fmaussion:
sorry for the noise. Seems to be related to certain shapefiles, some can be opened.

ocefpaf:
No problem.  The issue does exist because fiona is not 100% OK with `gdal 2.0.0`

Please try:

```shell
conda create -n conda_test python=2.7 geopandas
source activate conda_test
conda list | grep -E 'fiona|geopandas|gdal'
```

And you will see that you get:
```
fiona                     1.6.0               np110py27_0    defaults
geopandas                 0.1.1                    py27_1    ioos
libgdal                   2.0.0                         0    defaults
```

IMO the default channel is shipping a broken `fiona`.

The solution is to downgrade fiona, but that will bring all sort of pain due to numpy versions, etc. (Most packages are no longer compiled against numpy 1.9, forcing us to downgrade them all.)

This worked for me:

```shell
conda create -n conda_test python=2.7 geopandas numpy=1.9 gdal=1.11.2
```

[...]

PR #628, opened 2015-12-07
Added gdal and openjpeg

ocefpaf:
Building gdal first


ocefpaf:
`GDAL 1.11.3` is the first step to get #624 going.  Surprisingly I git it working on Windows.  However, I am clueless about the OSX failure.  The log is truncated and I do not own a Mac to test it :unamused: 

Ping @ChrisBarker-NOAA for some help!

ChrisBarker-NOAA:
has anyone tried this:

https://github.com/conda/conda-recipes/tree/master/gdal

it looks like it's set up for 1.11.2

Though in the end, maybe easier to fix Fiona :-)

ChrisBarker-NOAA:
Also -- any chance of using the libgdal already in the default conda channel, and only building the python bindings?

It's been a LONG time, but it used to be the case that the python bindings shipped with the GDAL lib sources, but building them was essentially an seconds step that you could run after the fact.

ocefpaf:
> has anyone tried this:
> https://github.com/conda/conda-recipes/tree/master/gdal
> it looks like it's set up for 1.11.2

Nope.  I am focusing on osgeo version to get ~~all~~ more of GDAL functionalities.

> Though in the end, maybe easier to fix Fiona :-)

Probably.  I see some activity to get Fiona working with `GDAL 2.0.0`, until then I will keep trying this PR.

> Also -- any chance of using the libgdal already in the default conda channel, and only building the python bindings?

Nope.  If I understand Fiona correctly does not use the python bindings and it needs `GDAL< 2.0.0`

ChrisBarker-NOAA:
still a bit confused -- conda default supplies libgdal 1.11.2 -- and presumably that's the version that was used for older fiona packages (which did work, yes?) and libgdal, without the python bindings, doesn't have a dependency on any particular python or numpy.

or does libgdal not include ogr -- which is what Fiona really needs.

Or maybe I'm really confused, and conda never did have the gdal packages we need, in any version.

ocefpaf:
> still a bit confused -- conda default supplies libgdal 1.11.2 -- and presumably that's the version that was used for older fiona packages (which did work, yes?)

Yes. If you are wiling to use `numpy=1.9`, `gdal=1.11.2`, and `fiona=1.5.1` OK.  But any update will break fiona.

> or does libgdal not include ogr -- which is what Fiona really needs.

I am getting this

```
undefined reference to `OGRGeometryFactory::transformWithOptions(OGRGeometry const*,
OGRCoordinateTransformation*, char**)'
```

When using libgdal=1.11.2 on Windows.

ChrisBarker-NOAA:
What a pain -- just tryin gto figure out the dependencies here. If I do conda info fiona, I get a lot of builds -- the recent ones are all dependent on libgdal 2.0.0 -- are all of those broken?

if you go back, there is:

fiona-1.5.1-np19py27_0 

Which is not that old (2015-05-29)

but it depends on gdal 1.11.2

what's interesting is that the newer builds depend on "libgdal", rather than gdal. I'm guessing here that gdal1.11.2 is the python bindings and lib all wrapped up in one, and they have since spilt the lib from the python bindings (assuming that "gdal" is the python bindings -- they really should call that pygdal or something...)

Anyway, so it looks like:

fiona depends on libgdal, python, and numpy
gdal(pygdal) depends on numpy and python and hdf and netcdf.

But: the older bulds dumped pygdal and libgdal in one package: gdal

Anyway if the fiona-1.5.1-np19py27_0  works, it was built against the gdal 1.11.0 package.

So that *should* work, with a newer python and numpy.

Can you tell that I am trying really, really hard to not have to build gdal itself :-)

ocefpaf:
> What a pain -- just trying to figure out the dependencies here. If I do conda info fiona, I get a lot of builds -- the recent ones are all dependent on libgdal 2.0.0 -- are all of those broken?

Yes.  They are broken.  See https://github.com/ioos/conda-recipes/issues/623#issue-119727170 and https://github.com/Toblerity/Fiona/issues/239#issuecomment-119239402.

> if you go back, there is:
fiona-1.5.1-np19py27_0 
Which is not that old (2015-05-29)
but it depends on gdal 1.11.2

That is OK for most people that knows how to manage envs (or use conda-execute :wink: @pelson).  Most users will get an updated, and broken, fiona either by an active or an accidental update.

> what's interesting is that the newer builds depend on "libgdal", rather than gdal. I'm guessing here that gdal1.11.2 is the python bindings and lib all wrapped up in one, and they have since spilt the lib from the python bindings (assuming that "gdal" is the python bindings -- they really should call that pygdal or something...)

Correct. `libgdal` should be everything (not only the libs) and `gdal` should be the python bindings.  I won't comment on those naming choices :unamused: 

> Anyway, so it looks like:
fiona depends on libgdal, python, and numpy
gdal(pygdal) depends on numpy and python and hdf and netcdf.
But: the older builds dumped pygdal and libgdal in one package: gdal
Anyway if the fiona-1.5.1-np19py27_0  works, it was built against the gdal 1.11.0 package.
So that *should* work, with a newer python and numpy.

In theory yes. I was unable to build fiona, for a newer numpy or older numpy, using the libgdal 1.11.2 available in the default channel thought.

I did have success building gdal 1.11.3 (complete package) for all numpy and python versions on Windows and Linux.  These binaries do work to build fiona on Linux.  I do have a hiccup on Windows that seems to be path related:

```shell
C:\Users\appveyor\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0\VC\Bin\link.exe /DLL /nologo /INCREMENTAL:NO /LIBPATH:C:\conda\envs\_build\Library\lib /LIBPATH:C:\conda\envs\_build\libs /LIBPATH:C:\conda\envs\_build\PCbuild /LIBPATH:C:\conda\envs\_build\PC\VS9.0 gdal_i.lib /EXPORT:init_transform build\temp.win32-2.7\Release\fiona/_transform.obj /OUT:build\lib.win32-2.7\fiona\_transform.pyd /IMPLIB:build\temp.win32-2.7\Release\fiona\_transform.lib /MANIFESTFILE:build\temp.win32-2.7\Release\fiona\_transform.pyd.manifest
LINK : fatal error LNK1181: cannot open input file 'gdal_i.lib'
error: command 'C:\\Users\\appveyor\\AppData\\Local\\Programs\\Common\\Microsoft\\Visual C++ for Python\\9.0\\VC\\Bin\\link.exe' failed with exit status 1181
 
C:\conda\conda-bld\work\Fiona-1.6.2>if errorlevel 1 exit 1 
```

If I can figure that out we need only to fix OSX. ( I am skipping OSX for now since I cannot even see the logs!)

ChrisBarker-NOAA:
>> if you go back, there is:
fiona-1.5.1-np19py27_0 
> That is OK for most people that knows how to manage envs 

I wasn't thinking we wanted to use it -- just hat presumable fiona successfully built against the gdal 1.11.2  conda package, so it _should_ be able to do so again. changing python and numpy versions shouldn't change that.

> In theory yes. I was unable to build fiona, for a newer numpy or older numpy, using the libgdal
> 1.11.2 available in the default channel thought.

Well darn -- makes you wonder how the heck continuum built it..

> I did have success building gdal 1.11.3 (complete package) 

does that mean including the python bindings in the build? It would be nice to not do that. Though I suppose to terrible, as we all want to use gdal with python anyway :-)

> If I can figure that out we need only to fix OSX.

I"ll try to take a quick look -- but i"ve been down this road years ago -- and I"m not looking forward to it!

Have you seen this stuff? http://www.kyngchaos.com/software/unixport that was really helpful in the past.

ocefpaf:
> Well darn -- makes you wonder how the heck continuum built it..

Some of the old binaries were re-packed versions of the Gohlke binaries.

> I"ll try to take a quick look -- but i"ve been down this road years ago -- and I"m not looking forward to it!

I just want to see the error to check if is something simple or the kind of "just drop it" problem. Travis-CI is truncation the log before the error :unamused: 

> Have you seen this stuff? http://www.kyngchaos.com/software/unixport that was really helpful in the past.

Thanks.  I will take a look.

ChrisBarker-NOAA:
Hmm -- I'm building on OS-X as we speak, but it looks like you got it to work - nice!

OOPS, maybe not -- onanaocnda .org there are no OS-X pacakges, so these must have passed 'cause os-x was turned off.

though I just built it on my machine, and presto! it seems to have worked!

===== gdal-1.11.3-np110py27_0 OK =====
+ ogrinfo sites.shp
INFO: Open of `sites.shp'
      using driver `ESRI Shapefile' successful.
1: sites (Point)
TEST END: gdal-1.11.3-np110py27_0
# If you want to upload this package to anaconda.org later, type:
#
# $ anaconda upload /Users/chris.barker/miniconda2/conda-bld/osx-64/gdal-1.11.3-np110py27_0.tar.bz2
#
# To have conda build upload to anaconda.org automatically, use
# $ conda config --set anaconda_upload yes

All I did was turn os-x compilation on , and run:

conda build --numpy=1.10 gdal

ocefpaf:
Thanks.  I suspect it is a Travi-CI thing then... hard to debug without the logs.

ChrisBarker-NOAA:
Maybe this needs a new issue, but:

Continuum has separated out the gdal libs (libgdal) and the python bindings (gdal). As of gdal 2.*

As we want Fiona built against gdal1.*, we went back to the old way of building the python bindings an libs together. But I think we've broken gdal-python:

Fiona doesn't use the python bindings (correct???) -- to what we should probably do is build a libgdal, without python, and have Fiona depend on that. That way, users could use the gdal 3.0 package to get python gdal functionality, and still get Fiona to work with the older gdal libs. This way, we've broken the gdal-python.

Hopefully, it should be possible to install libgdal1.* and libgdal2.* together and have them not stomp on each-other. (maybe not on Windows, with the horrendous convention of not putting version names in dlls...

And we may need to call the "new" libgdal package libgdal1 or something  , so that conda knows that you want both libgdal1 and libgdal2 at the same time.

ChrisBarker-NOAA:
If the CI server isn't building OS-X successfully, I should be able to run obvious-ci and build the OS-X packages.

ocefpaf:
> But I think we've broken gdal-python:

Conda updates are a mystery to me and I see libgdal/gdal being update/downgraded in a variety of combinations.  I'd rather force gdal 1.11.3 on everybody that wants a working version of fiona.  Updating the split package might still mess this up.

ocefpaf:
> If the CI server isn't building OS-X successfully, I should be able to run obvious-ci and build the OS-X packages.

To ensure a "clean" environment I would like to build this on Travis-CI.  Not sure how I will do that though :expressionless:

ChrisBarker-NOAA:
> I'd rather force gdal 1.11.3 on everybody that wants a working version of fiona. Updating the split package might still mess this up.

that's a fine idea -- but how do we do it? if someone does:

conda install fiona

they'll get gdal1.11.3 for everyting -- fine, I suppose.

but then they do an install of gdal directly:

conda install gdal

and it will probably upgrade it

and/or they install something else that depends on gdal, maybe even gdal >= 2, and that will break everything.

so a libgdal1.11.3 package could make this all cleaner.

ChrisBarker-NOAA:
> To ensure a "clean" environment I would like to build this on Travis-CI. Not sure how I will do that though.

I wonder if there is any way to tell travis not to truncate the logs....

There is probably something on my system that travis doesn't have, but how to know what it is?

-CHB

ocefpaf:
> and/or they install something else that depends on gdal, maybe even gdal >= 2, and that will break everything.

That problem will exist either way because conda does not implement the concept of `lib<package><ver>` correctly.  On my Linux machine I have `libgdal1` and `libgdal2` installed side-by-side because `libgdal{1,2}` package only the libs (`*.so.1.*` and `*.so.2.*` respectively).  While on conda `libgdal` is and the all but python bindings.

IMO we make this less messy by shipping gdal "whole" like the developers intended.  The real solution would be to break it into `libgdal`, `gdal`, `python-gdal`, `perl-gdal`, `ruby-gdal`, etc.

> There is probably something on my system that travis doesn't have, but how to know what it is?

I am afraid that only the paid version of Travis-CI would give us what we need.

ChrisBarker-NOAA:
> IMO we make this less messy by shipping gdal "whole" like the developers intended. 

that has actually been a source of pain for ages with GDAL.

> The real solution would be to break it into libgdal, gdal, python-gdal, perl-gdal, ruby-gdal, etc.

I think that is the way to go, and kinda what continuum has done with gdal2

so I'm still confused as to why we can't do a libgdal-1.*, maybe naming it libgdal1, so as not to conflict with the default libgdal. It would not include python, and fiona would use it.

or, we could get really ugly and ship gdal with fiona in one package -- that's the traditional way on windows :-)

-CHB

ocefpaf:
> I think that is the way to go, and kinda what continuum has done with gdal2

Take a look at the ["libgdal2"](http://anaconda.org/anaconda/libgdal/2.0.0/download/linux-64/libgdal-2.0.0-0.tar.bz2) and you will see what I am talking about.  It is **everything** but the python bindings.  That is not the correct way to split a package!

Maybe we could make a better split (`lib`, `gdal`, `python-gdal`), but I think that will be a source of confusion because it is even more different from the default channel.

> or, we could get really ugly and ship gdal with fiona in one package 

There are wheels for that and conda recipes wrapping those wheels.

Right now I think that shipping a complete gdal is probably the least confusing solution.

ChrisBarker-NOAA:
> Take a look at the "libgdal2" and you will see what I am talking about. It is everything but the python 
> bindings. That is not the correct way to split a package!

God no -- I wonder if that was done on purpose, or if someone was lazy :-)

> Maybe we could make a better split (lib, gdal, python-gdal), but I think that will be a source of confusion because it is even more different from the default channel.

sure, but what is the goal here? I thought it was to support Fiona -- at least until Fiona grows GDAL2 support. In that case, I think it's the best idea to build a gdal lib that is what Fiona needs, but DOES not provide the pygdal bindings -- i.e. does not overwrite what the latest "gdal" conda package provides.

Probably give it a different name if it's different than what the old gdal-1.11.2 package provided. Heck, maybe lib_gdal_for_fiona if we want to be really clear. Or "just_libgdal".

It should contain just the libs, with no language bindings, etc. It probably shouldn't even ship the command line utilities, so that they won't overwrite the one provided by gdal2

And maybe it would turn out to be useful for other stuff, too.

-CHB

ocefpaf:
> God no -- I wonder if that was done on purpose, or if someone was lazy :-)

Instead of doing something like that I'd rather have a single package.

PR #632, opened 2015-12-11
Adding cell_tree2d (pyugrid)

ocefpaf:
Ping @ChrisBarker-NOAA

### DIFF
--- a/cell_tree2d/bld.bat
+++ b/cell_tree2d/bld.bat
@@ -1,0 +1,2 @@
+"%PYTHON%" setup.py install --single-version-externally-managed --record record.txt
+if errorlevel 1 exit 1
--- a/cell_tree2d/meta.yaml
+++ b/cell_tree2d/meta.yaml
@@ -1,0 +1,35 @@
+package:
+    name: cell_tree2d
+    version: 0.1.1
+
+source:
+    fn: cell_tree2d-0.1.1-1.zip
+    url: https://pypi.python.org/packages/source/c/cell_tree2d/cell_tree2d-0.1.1-1.zip
+    md5: 06f95190c341e9fc6963dbd40274ef98
+
+build:
+    number: 0
+    script: python setup.py install --single-version-externally-managed --record record.txt
+
+requirements:
+    build:
+        - python
+        - setuptools
+        - cython
+        - numpy x.x
+    run:
+        - python
+        - numpy x.x
+
+test:
+    imports:
+        - cell_tree2d
+
+about:
+    home: https://github.com/NOAA-ORR-ERD/cell_tree2d
+    license: Public Domain
+    summary: CellTree data structure as described by Garth and Joy 2010
+
+extra:
+    recipe-maintainers:
+        - ocefpaf


ocefpaf:
@ChrisBarker-NOAA the build worked.  The only testing is the import though.  (We could use something more robust.)

Note that the tag says `v0.1.0` but [setup.py](https://github.com/NOAA-ORR-ERD/cell_tree2d/blob/master/setup.py#L30) says `0.0.1`.

I won't merge this yet because we will need this only when https://github.com/pyugrid/pyugrid/pull/111 gets merged.  I am planning to send a py3k fix before that BTW.

cell_tree2d/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install --single-version-externally-managed --record record.txt
```
ChrisBarker-NOAA: I"ve taken to using pip intsead:

pip install .

I'm pretty sure that pip injects setuptools and those flags in, but also ensures that the all the pip-friendly meta-data will be there, now ans in the future.

I **think** that's the "recommended" way to do it now.

(this will also work with packages that are using raw distutils...)

cell_tree2d/run_test.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+cd $SRC_DIR
+python setup.py test
```
ChrisBarker-NOAA: OK, now I need to see how you got teset into setup.py -- oddlyl, that's not obvious from the pytest docs....

cell_tree2d/build.sh
ocefpaf: > I've taken to using pip instead:
pip install .

Interesting.  I will try that.  Thanks!

ChrisBarker-NOAA:
note that:

setup.py test

should work, so that could be added to the recipe.

ocefpaf:
Thanks @ChrisBarker-NOAA! I will be submitting the recipe to conda-forge and we can change it there.

ChrisBarker-NOAA:
BTW, we have updated some stuff in cell_tree2d, but it's al clean up of the setup.py and the like, no API or functionality differences, so we're not pushiong a new version to PyPi yet. But if anything is ugly or broken in the build, we can push a new one -- so let us know rather than kludge around it.

ocefpaf:
> so let us know rather than kludge around it.

This PR was already merged and the package binaries are already making its way into http://anaconda.org/IOOS/cell_tree2d/files

ChrisBarker-NOAA:
awesome!

-CHB


On Fri, Mar 4, 2016 at 12:03 PM, Filipe <notifications@github.com> wrote:

> so let us know rather than kludge around it.
>
> This PR was already merged and the package binaries are already making its
> way into http://anaconda.org/IOOS/cell_tree2d/files
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ioos/conda-recipes/pull/632#issuecomment-192442403>.
>



-- 

Christopher Barker, Ph.D.
Oceanographer

Emergency Response Division
NOAA/NOS/OR&R            (206) 526-6959   voice
7600 Sand Point Way NE   (206) 526-6329   fax
Seattle, WA  98115       (206) 526-6317   main reception

Chris.Barker@noaa.gov

PR #634, opened 2015-12-12
Adding python-ternary

ocefpaf:
See https://github.com/marcharper/python-ternary#gallery

### DIFF
--- a/python-ternary/bld.bat
+++ b/python-ternary/bld.bat
@@ -1,0 +1,2 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
--- a/python-ternary/build.sh
+++ b/python-ternary/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install
--- a/python-ternary/meta.yaml
+++ b/python-ternary/meta.yaml
@@ -1,0 +1,26 @@
+package:
+    name: python-ternary
+    version: "1.0"
+
+source:
+    fn: python-ternary-1.0.tar.gz
+    url: https://pypi.python.org/packages/source/p/python-ternary/python-ternary-1.0.tar.gz
+    md5: b4afa32a9cb8b8c7156ba9f9f9b481f8
+
+build:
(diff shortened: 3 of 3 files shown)


emiliom:
How cool! I didn't even know about this package (I haven't done ternary
plots in a while). Looking forward to trying it out, eventually.
Thanks.

ocefpaf:
I had the need to make one recently (ping @juoceano) and I thought that the community might need it too.

(`python-ternary` is very flexible and the gallery is quite comprehensive.)

emiliom:
Heh. Ternary plots are used heavily in geochemistry, so I had a hunch she had had something to do with it.

ocefpaf:
Anything to take people away from things like [origin](http://www.originlab.com/index.aspx?go=Products/Origin/Graphing/Ternary) :stuck_out_tongue_winking_eye:

emiliom:
I never used Origin, but I saw their ads all the time! I did use a Matlab toolbox for ternary plots, way back, for my PhD thesis.

Does python-ternary do [Piper / diamond plots](http://hydrogeotools.blogspot.com/2013/11/piper-diagrams.html)? It doesn't look like it. Just curious; I've never had that need myself.

BTW, a decent FOSS (and Python-based) replacement for a GUI desktop plotting program might be [Veusz](http://home.gna.org/veusz/examples/), which also does ternary diagrams. I've played with it just a little bit. And [it looks like Veusz and python-ternary have been deemed to be the best options for ternary plotting in Python!](http://stackoverflow.com/a/31487036)

ocefpaf:
Thanks for the `veusz` tip.  I heard about it but never tried.  I went for `python-ternary` because of matplotlib, but `veusz` does look like a nicer alternative.

emiliom:
Yeah, python-ternary definitely seems like a better option as a python library to be used in programs/scripts, and hence much more compelling for the ioos conda channel. I *think* veusz doesn't even use matplotlib! But veusz as a FOSS desktop GUI (and scriptable, too, I think), is probably one of the best open-source options out there.

PR #654, opened 2015-12-16
Re add rasterio

ocefpaf:
The default channel stopped updating `rasterio` at `0.25`.  See https://github.com/ioos/conda-recipes/pull/651#issue-122487200  also.

### DIFF
--- a/rasterio/bld.bat
+++ b/rasterio/bld.bat
@@ -1,0 +1,2 @@
+"%PYTHON%" setup.py build_ext -I"%LIBRARY_INC%" -lgdal_i -L"%LIBRARY_LIB%" install --single-version-externally-managed --record record.txt
+if errorlevel 1 exit 1
--- a/rasterio/build.sh
+++ b/rasterio/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py build_ext -I$PREFIX/include -L$PREFIX/lib -lgdal install --single-version-externally-managed --record record.txt
--- a/rasterio/meta.yaml
+++ b/rasterio/meta.yaml
@@ -1,0 +1,41 @@
+package:
+    name: rasterio
+    version: "0.30"
+
+source:
+    fn: rasterio-0.30.0.tar.gz
+    url: https://pypi.python.org/packages/source/r/rasterio/rasterio-0.30.0.tar.gz
+    md5: d20553b1531397c76925c2df1a5d53d6
+
+build:
+    number: 0
+    skip: True  # [win and py35 or osx]
+
+    entry_points:
+        - rio = rasterio.rio.main:cli
+
+requirements:
+    build:
+        - python
+        - numpy x.x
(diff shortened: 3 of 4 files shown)


ocefpaf:
Changes
=======

0.30.0 (2015-11-16)
-------------------
- Added window utilities: get_data_window(), window_union(),
  window_intersection(), windows_intersect() (496, 506).
- Warn when an alpha band that might provide a dataset mask is shadowed by a
  nodata attribute (508, 523).
- IPython is not the default interpeter for rio-insp and the documentation 
  saying it is has been corrected (518).
- Guard against creating datasets with block sizes larger than the dataset
  width and height. Such datasets are semi-broken and are likely to be 
  mangled when read (521).
- Refactor of the `rasterio.features` tests (522).

0.29.0 (2015-10-22)
-------------------
- Fill masked arrays in rio-calc when using Numpy 1.10.x as well as with 1.8.x
  (500).
- When a raster dataset is not tiled, blockxsize and blockysize items are no
  longer included in its `profile` property. This prevents meaningless block
  size parameters from stripped, not tiled, datasets from being used when
  creating new datasets (503).

0.28.0 (2015-10-06)
-------------------
- Ensure that tools module is packaged (489, 490). The rio-merge command was
  broken in 0.27.0 and is restored to working order in version 0.28.0.
- Add `precision` keyword argument to `index()` method (492).

0.27.0 (2015-09-25)
-------------------
- Ensure local uniqueness of the rio-shapes feature ids (479).
- Surface compression and interleaving as dataset properties and in rio-info
  (481). In the module, these are enums (`enums.Compression` and 
  `enums.Interleaving`); the values of the enums correspond to GDAL terms
  (i.e, "DEFLATE") and the names are what surface in the CLI ("deflate").
- Change get_window() and DatasetReader.window() to return a window guaranteed
  to cover the input bounding box (464, 485).
- Bug fix for improperly computed transforms of output file in tools.merge and
  rio-merge (485).
- More robust determination of dataset nodata values. In particular, the 
  absence of a nodata value is much more clear: dataset.nodata should never
  return an out of range value when there is no nodata value, it should always
  return `None` (485).

0.26.0 (2015-08-11)
-------------------
- Add dependency on click-plugins, a new project that takes over the plugin
  duties formerly assigned to cligj (426).
- Change rio-stack's --photometric=RGB option to --rgb (429). Other 
  photometric interpretations should be assigned using the --co option 
  added in 0.25.0.
- Allow for ndarray-like objects (like xray arrays), not just numpy arrays,
  in warp() and elsewhere (436).
- Add --rgb flag to rio-convert (439).
- Fixed resampling algorithm enumeration bug (441).
- Colormap handling was made too strict in 0.24.1 and has been made more
  forgiving. Callers are now warned when alpha values will be ignored instead
  of receiving exceptions (444).
- Add a .gitignore (445). Better late than never!
- Add a checksum() method to base dataset class and checksums to output of
  rio-info (449).

Issue #690, opened 2016-01-06
Package request


ocefpaf:
@efiring we have a few requests to provide conda binaries for [CODAS](http://currents.soest.hawaii.edu/docs/adcp_doc/index.html).

I would like to know your opinion before I try anything. (It has been a while since the last time I had to install the whole CODAS and UHDAS stack.)

efiring:
Interesting.  The idea has been in the back of my mind for a while, but I haven't tried anything or thought about it in detail.  Installing into an existing conda environment is easy, at least on linux and OS X.  It has also been done on Windows 7 and maybe 8, but not 10, as far as I know.  I don't know whether the present conda mingw toolchain works on 10.
In any case, if you want to try it, versions of installation instructions are in http://currents.soest.hawaii.edu/docs/adcp_doc/codas_setup/index.html and http://currents.soest.hawaii.edu/ocn_data_analysis/.  In addition to questions about what can be done in various Windows versions there is the wrinkle that our GUI tools for SADCP processing use wxpython, which is not yet available for Python 3.  Apart from that, everything we have should run on 2.7 or 3.4+.

ocefpaf:
Thanks @efiring!  I see that the installation is much easier than before.  I will take a shot at it and keep you posted.

ocefpaf:
Closing this as `codas` and `pycurrents` are in the repo.

PR #697, opened 2016-01-11
codas

ocefpaf:
**This PR is not ready yet!  Do not merge.** (See #690)

@efiring do you plan to tag releases for these or should we just use `tip` for now?

TODO:
- [X] be smarter about the dependencies `run`/`build`;
- [X] does `CODAS3` depends on python at all?  If not we don't need to build it for `PY{27,34,35}`
- [ ] add more tests (execute a command or something that might trigger an error);
- [ ] polish the packages metadata: version, license,  summary;
- [ ] add the "offline" steps: bathymetry, ?
- [ ] move instead of copying the `.exe` from `bin` to `%Scripts%` to avoid bloating the binary.
- [ ] add a [notebook](https://github.com/ioos/conda-recipes/tree/master/notebooks) testing this with a known dataset.

Codas still builds3 times (for python 2.7, 3.4, and 3.5) because of the build dependency on python.  I need to check if there is a conda-build special arg to avoid that.

    codas-3.0.dev-0 (will be built: True)
    codas-3.0.dev-0 (will be built: True)
    codas-3.0.dev-0 (will be built: True)


ocefpaf:
Conda is finding the previous `tip.tar.gz` file in the cache:

    Found source in cache: tip.tar.gz

That breaks the build.  It would be nice to get named tag releases.

efiring:
I suspect we should move to a tag scheme but I will have to discuss this with Jules and Toby.
I don't think that uhdas and onship should be included at all for now, though; they are completely specific to setting up ADCP logging on known ships, and they are only a part of the rather involved process required to do this.  Longer term, we might very well want to use conda for some of the installation--I was already thinking about it--but even then I suspect it would not be appropriate for uhdas and onship to be hosted by ioos.

ocefpaf:
> I suspect we should move to a tag scheme but I will have to discuss this with Jules and Toby.

That would be awesome.

> I don't think that uhdas and onship should be included at all for now

OK. I just removed them.

> but even then I suspect it would not be appropriate for uhdas and onship to be hosted by ioos.

OK.  But when the time comes consider hosting the recipes for `uhdas` and `onship` at [conda-forge](https://github.com/conda-forge). There are many advantage of hosting recipes in a community driven channel/repository.

ocefpaf:
The Linux failure is due to the lack of mercurial in our Docker image.  That will be solved soon, but maybe we could patch `pycurrents` to avoid having mercurial as a dependency.

The Windows failure is more complicated.  I have no idea what

    Checking for code snippet                : no

means.

efiring:
It is not supposed to require hg; this is a bug that was pointed out in another context a few days ago.  I will fix it shortly.

efiring:
Windows: it looks like the problem might be the presence of msvc, which waf is giving priority to over the gcc from mingw32 in Anaconda.

efiring:
For Windows, try adding an option to the waf configure and build line:
```
waf --check-c-compiler=gcc configure configure build
```
An updated waf might be needed:  http://stackoverflow.com/questions/14072705/waf-conflict-between-mingw-gcc-and-msvc-2012
If so, I can do that.

ocefpaf:
>  It is not supposed to require hg; this is a bug that was pointed out in another context a few days ago. I will fix it shortly.

Thanks!  (Any change of moving to GitHub BTW?  I could help with that.)

> An updated waf might be needed: http://stackoverflow.com/questions/14072705/waf-conflict-between-mingw-gcc-and-msvc-2012
If so, I can do that.

No need. `Codas` is building just fine now. However, `pycurrents` is having trouble locating `codas` binaries.  That should be easy to fix. I will take a better look tomorrow.

PS: @efiring I am naming the package `codas` version 3 rather than `codas3`,
is that correct?

efiring:
On 2016/01/11 1:37 PM, Filipe wrote:
> (Any change of moving to GitHub BTW? I could help with that.)

We've started considering it.  I have mixed feelings about that; in some 
ways Mercurial is better suited for the way Jules and I work.

efiring:
On 2016/01/11 1:37 PM, Filipe wrote:
> PS: @efiring <https://github.com/efiring> I am naming the package
> |codas| version 3 rather than |codas3|,
> is that correct?

I think that is OK.  We have never really used versioning in the normal 
sense.  It has been just "codas3" for as long as it has been in general 
use; but if we start tagging releases, then there is no reason they 
couldn't be 3.1, 3.2, etc.  That codebase is pretty stable, though, 
fortunately.

ocefpaf:
> We've started considering it.  I have mixed feelings about that; in some
ways Mercurial is better suited for the way Jules and I work.

I understand.  (I had a hard time moving everything to Git and GitHub.)

I believe I fixed `codas` path and `pycurrents` is able to find the binaries on Windows now.

Now I am stuck with another issue:

```shell
LINK : fatal error LNK1181: cannot open input file 'codas3_static.lib' 
error: Command "C:\Program Files (x86)\Microsoft Visual Studio 9.0\VC\BIN\link.exe /DLL /nologo /INCREMENTAL:NO /LIBPATH:C:\conda\envs\_build\lib /LIBPATH:C:\conda\envs\_build\libs /LIBPATH:C:\conda\envs\_build\PCbuild /LIBPATH:C:\conda\envs\_build\PC\VS9.0 codas3_static.lib /EXPORT:init_codas build\temp.win32-2.7\Release\codas/src/_codas.obj build\temp.win32-2.7\Release\codas/src/getadcp.obj /OUT:build\lib.win32-2.7\pycurrents\codas\_codas.pyd /IMPLIB:build\temp.win32-2.7\Release\codas/src\_codas.lib /MANIFESTFILE:build\temp.win32-2.7\Release\codas/src\_codas.pyd.manifest" failed with exit status 1181
```

efiring:
This looks like the problem with msvc getting in the way again.  The error is coming from Visual Studio, which never should be involved.  It looks like setup in pycurrents is trying to use msvc instead of gcc.  I don't know how to fix that short of blowing away Visual Studio, but there must be a way.

ocefpaf:
> This looks like the problem with msvc getting in the way again. 

Yep.

>  I don't know how to fix that short of blowing away Visual Studio, but there must be a way.

We could try using `setuptools` and specifying the compiler.

I will test the Linux package first, to check if we are going in the right direction, and then come back to this.

efiring:
Alternative: figure out how to build everything with Visual Studio.  In principle this shouldn't be too hard for someone with a VS setup, but we don't have that--and I really don't want to plunge into the MS compiler mess myself in any case.  It's probably the best way to solve the problem, at least as an option, given all the present problems with mingw-related approaches.  I gather from the numpy list that a mingw64-based solution might be close, but there is no telling how long it will actually be before it is fully operational and packaged as part of Anaconda--if ever.

rsignell-usgs:
That's the beauty of building on appveyor:  all the MS compilers and tools are already there!

efiring:
Right, but someone still has to figure out what needs to be changed in my waf wscript for codas (maybe just the specification of mingw in the Win environment) and in my setup.py for the python modules, and then come up with the conda batch file.

ocefpaf:
> That's the beauty of building on appveyor: all the MS compilers and tools are already there!

Indeed. And that did helped us in a number of other packages here.  However, we rarely have those compiler locally to test the recipes before submitting a PR. (I don't even have a Windows machine!)  That makes creating new packages for Windows trial-and-error game with AppVeyor.

> I gather from the numpy list that a mingw64-based solution might be close, but there is no telling how long it will actually be before it is fully operational and packaged as part of Anaconda--if ever.

I do have high hopes for that project I and strongly believe it will be available in conda.  I follow [carlkl's](https://anaconda.org/carlkl/mingwpy) conda channel very closely. Even used his current version of mingwpy in a few Fortran projects here.

> Right, but someone still has to figure out what needs to be changed in my waf wscript for codas (maybe just the specification of mingw in the Win environment) and in my setup.py for the python modules, and then come up with the conda batch file.

I do have some idea, just out not enough time  :unamused:
Maybe I can removed the Windows builds for now and merge the Linux and OS X packages to get people going.

ocefpaf:
@efiring things are "compiling" now but we need some real testing. (Ping @saulomeirelles wanna try it?)

I am using `mingwpy` for `pycurrents` and conda's `mingw` for `codas`.  Maybe that will break things, but I was unable to get `pycurrents` to compile with `mingw`. (Missing `libpython27.a` error...  I will re-try it and post the full error later.)  I also do not understand `waf` enough to compile codas with `mingwpy`, hence the mismatch. But I believe it is possible, so I will look into that later.

Note that I had to skip the build on Windows+Python 3 due to the dependency on `mercurial`.  (If I get some time I'll send a patch your way to fix that, as you said above there is no reason to look for `hg` at build time.)

efiring:
@ocefpaf All of that copying should not be necessary for codas3.  What do the 3 variables resolve to?  (LIBRARY_INC, LIBRARY_LIB, SCRIPTS)
I have made some changes to try to make msvc compilation work with Anaconda, but they should work for gcc also.  I'm not actually sure to what extent the present state works with msvc, since I have very little testing capability, and there are all the different versions.  In any case, I would like to get codas3/wscript in a state that works for everything possible, and without a lot of extra lines required in the recipe.
I have also fixed the missing-hg problem in pycurrents, I think.
Newest versions are in http://currents.soest.hawaii.edu/hgstage.

saulomeirelles:
Great news! I'll give a try over the weekend.

ocefpaf:
> @ocefpaf All of that copying should not be necessary for codas3. What do the 3 variables resolve to? (LIBRARY_INC, LIBRARY_LIB, SCRIPTS)

Maybe some of them are not needed (and I could probably move, using glob syntax, instead of coping too).  I confess I did the copying until things "worked" without a proper test.  But we do need some of it because of the way conda handle the paths on Windows. Some of the binaries and DLLs are not found when creating conda envs unless they are in the `Scripts` directory. (For example see the [udunits2](https://github.com/SciTools/conda-recipes-scitools/blob/8ee40bc286a203f6649c8a27a1b7ad5028b68299/udunits2/bld.bat#L20-L21) recipe.)

BTW they resolve to:

```
LIBRARY_INC: <build prefix>\Library\include
LIBRARY_LIB: <build prefix>\Library\lib
SCRIPTS: <build prefix>\Scripts
```

(See http://conda.pydata.org/docs/building/environment-vars.html for more info)

saulomeirelles:
@ocefpaf I am still running into troubles. Check it out:

ME_CE_20150125_0134_ADCP001_000000.STA  ME_CE_20150131_0549_ADCP000_000006.ENR
ME_CE_20150125_0134_ADCP001_000000.VMO  ME_CE_20150131_0549_ADCP000_000006.ENS
ME_CE_20150125_0134_ADCP002_000000.ENS  ME_CE_20150131_0549_ADCP000_000006.ENX
ME_CE_20150125_0134_ADCP002_000000.ENX  ME_CE_20150131_0549_ADCP000_000007.ENS
ME_CE_20150125_0134_ADCP002_000000.LOG  ME_CE_20150131_0549_ADCP000_000008.ENR
ME_CE_20150125_0134_ADCP002_000000.LTA  ME_CE_20150131_0549_ADCP000_000008.ENX
saulo@smnr:~/Google Drive/ADCP_RAW/ADCP_Janeiro_2015$ quick_adcp.py 
Traceback (most recent call last):
  File "/usr/bin/quick_adcp.py", line 3, in <module>
    from pycurrents.adcp.quick_adcp import quick_adcp
  File "/usr/local/lib/python2.7/dist-packages/pycurrents/__init__.py", line 8, in <module>
    from future import standard_library
ImportError: No module named future
saulo@smnr:~/Google Drive/ADCP_RAW/ADCP_Janeiro_2015$ which python
/home/saulo/miniconda/bin/python
saulo@smnr:~/Google Drive/ADCP_RAW/ADCP_Janeiro_2015$ conda install future
Fetching package metadata: ......
Solving package specifications: ...................
# All requested packages already installed.
# packages in environment at /home/saulo/miniconda:
#
future                    0.15.2                   py27_0    defaults
saulo@smnr:~/Google Drive/ADCP_RAW/ADCP_Janeiro_2015$ 

I installed codas via conda: conda install codas. 

Any thought? 

Cheers, Saulo

ocefpaf:
Can you send us the results of `which quick_adcp.py ` (o locate `quick_adcp.py`).

If I remember codas correctly that is the processing script, right?  I believe you might have another installation there because `future` is listed and installed in the conda package, but I am getting an error too when testing `quick_adcp.py`:

```shell
Traceback (most recent call last):
  File "/home/filipe/miniconda3/envs/IOOS/bin/quick_adcp.py", line 3, in <module>
    from pycurrents.adcp.quick_adcp import quick_adcp
  File "/home/filipe/miniconda3/envs/IOOS/lib/python2.7/site-packages/pycurrents/adcp/quick_adcp.py", line 117, in <module>
    from pycurrents.adcp.quick_setup    import Q_setup
  File "/home/filipe/miniconda3/envs/IOOS/lib/python2.7/site-packages/pycurrents/adcp/quick_setup.py", line 24, in <module>
    from pycurrents.adcp.uhdasconfig import UhdasConfig
  File "/home/filipe/miniconda3/envs/IOOS/lib/python2.7/site-packages/pycurrents/adcp/uhdasconfig.py", line 35, in <module>
    import uhdas.uhdas.rdi_setup as rdi
ImportError: No module named uhdas.uhdas.rdi_setup
``` 

The `uhdas` module was not packaged (see https://github.com/ioos/conda-recipes/pull/697#issuecomment-170627587).

@efiring do we have to package `uhdas` or is there a way to run it without it?

PS: It would be nice to run a very simple test with `quick_adcp.py` and some real data at the end of the build.  I will try to put one together, but appreciate someone can point me to a free, simple, and small dataset.

saulomeirelles:
Thanks for your support, @ocefpaf !

Indeed, after getting CODAS correctly installed, I got same error as you:

(CODAS)saulo@smnr:~/Google Drive/ADCP_RAW/ADCP_Janeiro_2015$ quick_adcp.py 
Traceback (most recent call last):
  File "/home/saulo/miniconda/envs/CODAS/bin/quick_adcp.py", line 3, in <module>
    from pycurrents.adcp.quick_adcp import quick_adcp
  File "/home/saulo/miniconda/envs/CODAS/lib/python2.7/site-packages/pycurrents/adcp/quick_adcp.py", line 117, in <module>
    from pycurrents.adcp.quick_setup    import Q_setup
  File "/home/saulo/miniconda/envs/CODAS/lib/python2.7/site-packages/pycurrents/adcp/quick_setup.py", line 24, in <module>
    from pycurrents.adcp.uhdasconfig import UhdasConfig
  File "/home/saulo/miniconda/envs/CODAS/lib/python2.7/site-packages/pycurrents/adcp/uhdasconfig.py", line 35, in <module>
    import uhdas.uhdas.rdi_setup as rdi
ImportError: No module named uhdas.uhdas.rdi_setup

ocefpaf:
We are working on that.  Follow the progress on #751.  Once that gets merged I believe we'll have a useful version of the package.

PR #705, opened 2016-01-13
use pip to install

ChrisBarker-NOAA:
I like the pip command better -- certainly easier to remember. And I found the setup.py install method didn't work fo rme -- it couldn't find the source datafile to make the db. whereas pip did work.

As for:
```
# FIXME: Do we need a post-link.sh with:
# $PYTHON setup.py remake_oil_db
```
no -- that's not necessary -- the install should build the db fresh
 -- remake_oil_db is to force rebuilding the db if the source data has changed.

### DIFF
--- a/oil_library/build.sh
+++ b/oil_library/build.sh
@@ -1,10 +1,6 @@
 #!/bin/bash
 
 pushd oil_library
-$PYTHON setup.py install  --single-version-externally-managed --record record.txt
+$PYTHON pip install ./
 popd
 
-# FIXME: Do we need a post-link.sh with:
-# pushd $RECIPE_DIR
-# $PYTHON setup.py remake_oil_db
-# popd


ocefpaf:
> And I found the setup.py install method didn't work for me

Interesting.  I do like the short command!  How did `setup.py install` failed in you case?
@pelson is there a reason to not use pip?

I do like the fact that `--single-version-externally-managed` won't try download the run dependencies at build time.  I believe there is an equivalent for pip, right?

PS: You need to bump the build [number](https://github.com/ChrisBarker-NOAA/conda-recipes/blob/patch-1/oil_library/meta.yaml#L10).

oil_library/build.sh
```diff
@@ -1,10 +1,6 @@
 #!/bin/bash
 
 pushd oil_library
-$PYTHON setup.py install  --single-version-externally-managed --record record.txt
+$PYTHON pip install ./
```
ChrisBarker-NOAA: maybe add --nodeps here...

ocefpaf:
I think it is best if you made the NOAA-ORR conda-channel the canonical place for this recipe.  I will add an `expand source` script pointing there in another PR.

ChrisBarker-NOAA:
for the record, you can do:

pip install --no-deps ./

But yes, let's make the NOAA-ORR channel the canonical channel for this one. I think that recipe and build is working fine now.

py_gnome is a different story, not sure if we got that fixed yet. James was working on it.

PR #706, opened 2016-01-13
post_gnome

ocefpaf:
Adding post_gnome.

Ping @rsignell-usgs.  Not sure if the package will work.  I had to patch a few things just to get it to build.


ocefpaf:
We can drop the patch once PRs https://github.com/NOAA-ORR-ERD/GnomeTools/pull/2 and https://github.com/NOAA-ORR-ERD/GnomeTools/pull/3 are merged.

PR #722, opened 2016-01-20
Added pynio and dependencies.

ocefpaf:
This will be an optional  dependency of the next `xarray ` release.

(See https://github.com/NCAR/pynio.git.)


ocefpaf:
For linux we will have to wait until https://github.com/conda/conda-build/pull/731 makes in the a next release.  For OS X I have no idea where to start.

pynio/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+        - hdf5
+        - hdf4
+        - proj.4
+        - gdal
+        - curl
+        - libiconv
+        - g2clib
+        - hdfeos2
+        - hdfeos5
+        - libgcc
```
ocefpaf: Ping @bekozi and @pelson.

I am starting to think that we should move to conda's `gcc` and `libgcc` to avoid relying too much on pre-installed on system libraries.  What do you think?

Issue #741, opened 2016-02-02
update ulmo recipe for new release (v0.8.2)


dharhas:
@ocefpaf 

Hey we just did a point release for ulmo with a few bugfixes and an additional data source. Can you update the conda package on the ioos channel. Thanks.

ocefpaf:
Done in #742 

@dharhas we are limiting `pytables` to `<=3.1.1` on Windows and Python 3, but that version is not available for Python 3.5.  Is that condition still true? (I don't remember why we need that version.)

dharhas:
@ocefpaf pytables 3.2.0 had a hard crash on python 3.4/windows which is why we had the limit. They may have fixed it by now.

PR #742, opened 2016-02-02
ulmo 0.8.2

ocefpaf:
Changelog
---------
**0.8.2 (released 2016-02-02)**

- added modules for getting LCRA hydromet and water quality data
- fixed waterml parser bug that was affecting USGS NWIS data when
  parameter data collected using multiple methods exist (e.g. water level
  measured with RADAR and pressure transducer) through use of optional 'method'
  kwarg
- fixed ncdc.ghcn.get_stations() failing tests with pandas==0.17.0

### DIFF
--- a/ulmo/meta.yaml
+++ b/ulmo/meta.yaml
@@ -1,15 +1,14 @@
 package:
     name: ulmo
-    version: "0.8.1"
+    version: "0.8.2"
 
 source:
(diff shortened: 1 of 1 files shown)


ocefpaf:
Fixes #741

ocefpaf:
Thanks @dharhas.  Not that I release the `pytables` version on Windows.  Let me know if you find any problems with that.

PR #764, opened 2016-02-15
Bump luigi to 2.0.1 (windows support)

kwilcox:
https://github.com/spotify/luigi/releases/tag/2.0.1


ocefpaf:
@kwilcox I think you are using `ioos/conda-recipes` repository for branches instead of `kwilcox/coda-recipes`, rught? That results in the recipes being built and uploaded before the PR is merged, which is not ideal!

See http://anaconda.org/IOOS/luigi/files

ocefpaf:
@kwilcox we need a `bld.bat` to build on Windows, but `luigi` depends on `python-daemon` which is not available on Windows, is that correct?

Unless I am missing something here let's use #766 instead.

Issue #723, opened 2016-01-21
nco needs libgomp1


daf:
I was building a Docker image with conda and the ioos channel, and installed nco (4.5).  None of the binaries would work, giving this:

```
root@008a7db6d797:/data# ncks
ncks: error while loading shared libraries: libgomp.so.1: cannot open shared object file: No such file or directory
```

Ended up solving it by adding an `libgomp1` line to my list of things to `apt-get install` in the Dockerfile, but is this something that can be avoided/tested/checked so others aren't so confused as I was?

ocefpaf:
We need to document that.  @bekozi already did a nice work of tracking some these down in #676.

I just need to update the wiki... :disappointed:

ocefpaf:
Fixed in #770

Issue #700, opened 2016-01-12
Possible CDO linking issues


bekozi:
I encountered some linking issues with ``cdo`` and Ubuntu. Let me know if I can help out with any more debugging.

``libgrib`` not linking (probably also requires ``build-essential``):
```
$ docker run continuumio/anaconda bash -c "conda install --yes -c ioos cdo python=2.7 && cdo"
cdo: error while loading shared libraries: libgrib_api.so.0: cannot open shared object file: No such file or directory
```

Requires ``build-essential``:
```
$ docker run continuumio/anaconda bash -c "conda install -c ioos cdo python=3 && cdo"
cdo: error while loading shared libraries: libgomp.so.1: cannot open shared object file: No such file or directory
```

This works:
```
$ docker run continuumio/anaconda bash -c "apt-get update --yes && \
                                           apt-get install --yes build-essential && \
                                           conda install -c ioos cdo python=3 && cdo"
```

This line looks a little suspect for ``ecmwf_grib``: https://github.com/SciTools/conda-recipes-scitools/blob/master/ecmwf_grib/meta.yaml#L17. However, I would think that would be problematic for the Python 3 builds not 2.7.

ocefpaf:
I was aware of the `grib_api` issue and I have been looking the other way for a while :unamused:

My solution?  Get rid of it :wink: (See PR #702.)
Most of our users do not use the `grib_api`.  I will re-try that in the future though.

Regarding the `build-essential` dependency: that should go in https://github.com/ioos/conda-recipes/issues/676#issuecomment-167015494.

I will edit the wiki and we will ask @rsignell-usgs to review it later.  (Let's keep this issue open until I do so.)

Thanks for looking into this @bekozi!

bekozi:
> My solution? Get rid of it (See PR #702.)

A fine solution. :smile: 

> Regarding the build-essential dependency: that should go in #676 (comment).

Sounds good!

ocefpaf:
Fixed in #770

Issue #676, opened 2015-12-21
Required system libraries for packages


bekozi:
What is the best practice for indicating required system libraries for a conda package?

For instance, when installing ``ESMF`` on Ubuntu, ``build-essential`` and ``gfortran`` are required (i.e. ``apt-get install build-essential gfortran``). Also, ``libexpat1`` is required for ``udunits2``. I am thinking ``expat`` should maybe be a requirement for ``udunits2``?

Thanks!

ocefpaf:
That is a good question!  We need to make that clearer somehow...

Right now there is no clear way to declare them for all CIs.  We add them separately in the [`.travis.yml`](https://github.com/ioos/conda-recipes/blob/master/.travis.yml#L26-L33) for OS X and the [`Dockerfile`](https://github.com/ioos/conda-recipes/blob/master/Dockerfile#L7-L41) for Linux. On Windows we must package them!  (See the [`expat`](https://github.com/SciTools/conda-recipes-scitools/tree/master/expat) and the [`udunits2`](https://github.com/SciTools/conda-recipes-scitools/blob/master/udunits2/meta.yaml#L10-L14) recipes for an example.)

bekozi:
I see! It would be nice to have platform pages where recommended packages could be listed for ``yum`` /  ``apt-get`` / ``brew``, etc. Package-specific might be too much to manage, but a list of recommended Ubuntu packages (for example) seems reasonable. Could sections be added in the "Setting up the IOOS Python Environment for..." pages?

ocefpaf:
> package-specific might be too much to manage

It is indeed but necessary for build reproducibility.

> Could sections be added in the "Setting up the IOOS Python Environment for..."

We can certainly improve the docs, but ideally no system library is needed to use the packages (only to build them).  I am trying to eliminate that problem when it crops up.  (See #667 for example.)

bekozi:
> We can certainly improve the docs, but ideally no system library is needed to use the packages (only to build them).

Ahhh, okay. Thanks for the clarification. So for ``ESMF`` this installation test fails:

```
docker run continuumio/miniconda bash -c 'conda install -c ioos esmpy && python -c "import ESMF; m = ESMF.Manager(); print(m.pet_count)"'

Traceback (most recent call last):
  File "/opt/conda/lib/python2.7/site-packages/ESMF/interface/loadESMF.py", line 111, in <module>
    mode=ct.RTLD_GLOBAL)
[...]

PR #769, opened 2016-02-17
Fix libgomp

ocefpaf:
I am not sure if this is the way to go.  Ping @daf and @bekozi who reported #723 and #700 respectively.

I tested using my docker image by uninstalling all the compiler and related system libraries and then built the packages with the default channel's `gcc`.  It worked locally, but I am not sure if it will work using our Docker image. (If not I am considering removing all the compiler from that image too and re-build everything using conda's gcc.)

### DIFF
--- a/cdo/meta.yaml
+++ b/cdo/meta.yaml
@@ -17,11 +17,13 @@ requirements:
         - libnetcdf
         - netcdf-fortran
         - proj.4
+        - gcc
     run:
         - jasper
         - libnetcdf
         - netcdf-fortran
         - proj.4
+        - libgcc
 
 test:
     commands:
--- a/nco/meta.yaml
+++ b/nco/meta.yaml
@@ -19,11 +19,13 @@ requirements:
         - hdf5
         - libnetcdf
         - udunits2
+        - gcc
     run:
         - gsl
         - hdf5
         - libnetcdf
         - udunits2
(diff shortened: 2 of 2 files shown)


ocefpaf:
Ouch!  I forgot to bump the build number... Sending another PR!

bekozi:
> I tested using my docker image by uninstalling all the compiler and related system libraries and then built the packages with the default channel's gcc. It worked locally, but I am not sure if it will work using our Docker image.

Thanks for the ping. I think this should generally work, but there may be complications with some default channel packages (i.e. https://github.com/ioos/conda-recipes/issues/676#issuecomment-166982432). It would be great to move to the default channel ``gcc`` eventually to avoid having users install ``gfortran`` for instance.

FYI, I am doing some basic testing for ``ESMF`` and ``ocgis`` installed from IOOS with "clean" docker images: https://github.com/NESII/conda-recipes-ioos/blob/test-install-extras/.travis.yml.

ocefpaf:
>  It would be great to move to the default channel gcc eventually to avoid having users install gfortran for instance.

Yeah, that is the goal.  BTW I am testing these two here and, if that work, I will be sending them to conda-forge.  I will keep you posted as, eventually, I will send ESMF to conda-forge too.  Let's just work out the kinks first.

PR #774, opened 2016-02-29
Add codar2netcdf

kwilcox:
A utility to convert CODAR total ASCII files into CF NetCDF files


ocefpaf:
@kwilcox the failure has nothing to do with the package. I will fix it soon.

PR #777, opened 2016-03-01
Point to the feedstock.

ocefpaf:
Should fix the failures in our CIs.

### DIFF
--- a/netcdf4/source.txt
+++ b/netcdf4/source.txt
@@ -1,3 +1,3 @@
-gh-repo: conda-forge/staged-recipes
+gh-repo: conda-forge/netcdf4-feedstock
 git-tag: master
-contents: recipes/netcdf4/*
+contents: recipe/*


pelson:
Hmmm. Not sure what's going on here.

ocefpaf:
I cannot see the problem either.  (And if I cannot find it I'll just remove netcdf4 from IOOS and start educating our people on conda-forge :smile:)

ocefpaf:
Found the problem.  The mystery is how `AppVeyor` passed!

pelson:
What was it? I'm curious.

netcdf4/source.txt
```diff
@@ -1,3 +1,3 @@
-gh-repo: conda-forge/staged-recipes
+gh-repo: conda-forge/netcdf4-feedstock
 git-tag: master
-contents: recipes/netcdf4/*
+contents: recipe/*
```
pelson: Ah, was this still netcdf4?

ocefpaf: Yep.  That was my mistake.

PR #779, opened 2016-03-02
Update compliance-checker to 2.1.0

daf:
(plz test)

### DIFF
--- a/compliance-checker/meta.yaml
+++ b/compliance-checker/meta.yaml
@@ -15,7 +15,7 @@ build:
 
 requirements:
     build:
-        - python >=2.7
+        - python
         - setuptools
     run:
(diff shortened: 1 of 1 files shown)


compliance-checker/meta.yaml
```diff
@@ -15,7 +15,7 @@ build:
 
 requirements:
     build:
-        - python >=2.7,<3
+        - python >=2.7
```
ocefpaf: You can remove the ` >=2.7` as well. We only build  2.7, 3.4, and 3.5.

ocefpaf:
@daf The self contained CIs do not seems to suffer from the `cf-units` vs `cf_units` issue.

PR #795, opened 2016-03-03
pygeogrids

ocefpaf:
See https://github.com/TUW-GEO/pygeogrids

Ping @rsignell-usgs and @ChrisBarker-NOAA this package is for Regular grids what `pysgrid` and `pyugrid` are (should be?) for SGRID and UGRID.

### DIFF
--- a/pygeogrids/bld.bat
+++ b/pygeogrids/bld.bat
@@ -1,0 +1,2 @@
+"%PYTHON%" setup.py install --single-version-externally-managed --record record.txt
+if errorlevel 1 exit 1
--- a/pygeogrids/meta.yaml
+++ b/pygeogrids/meta.yaml
@@ -1,0 +1,35 @@
+package:
+    name: pygeogrids
+    version: 0.1.9
+
+source:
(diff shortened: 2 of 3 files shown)


ChrisBarker-NOAA:
Cool! I"ll need to take a closer look at this. Bu tthe idea ties into what I"d been thinking we really should do -- get away from py_ugrid, py_sgrid, etc, and have a meta-package: py_star_grid (*grid, i.e. any grid, get it?)

The idea would be that a user could open up a data set defined on any (supported) grid, and have a single API to work with -- interpolate, plot, etc, etc....

Maybe pygeogrids can provide the regular gridded version.

More on this somewhere else -- the APIRUS project?

ocefpaf:
> More on this somewhere else -- the APIRUS project?

:+1:

ChrisBarker-NOAA:
Hmm, a quick look shows:


"Nearest neighbor search"
"conversion of grid piont indices to lat, lon"
and:
"pykdtree"

for regular grids, you don't need a fancy spatial index to find where you are in the grid -- what's going on here?

-CHB

Issue #776, opened 2016-03-01
nco and cdo errors


DamienIrving:
I've installed [nco](http://anaconda.org/ioos/nco) and [cdo](http://anaconda.org/ioos/cdo) on OS X (version 10.11) using the relevant IOOS packages but then get the following errors when I try and run them from the command line:
```
$ ncks
dyld: Library not loaded: @rpath/libgssapi_krb5.2.2.dylib
  Referenced from: /Users/irv033/miniconda2/envs/IOOS/bin/ncks
  Reason: image not found
Trace/BPT trap: 5

$ cdo
dyld: Library not loaded: /usr/local/lib/libuuid.16.dylib
  Referenced from: /Users/irv033/miniconda2/envs/IOOS/bin/cdo
  Reason: image not found
Trace/BPT trap: 5
```
(It looks like this issue has been discussed in the past [here](https://github.com/Anaconda-Server/docs.anaconda.org/issues/98), however @CloLanglais informs me that she isn't sure if that solution works with the latest version of OS X)

What would be the best solution to this problem and would it be possible to update http://anaconda.org/ioos/nco and http://anaconda.org/ioos/cdo accordingly?

kwilcox:
I don't have OS X to test with. Can you try `conda install krb5` (install from the default channel) and see if that clears up the `ncks` issue?

DamienIrving:
@kwilcox Yep, `conda install krb5` fixes the issue with any/all of the nco operators (like `ncks`). Thanks!

It doesn't fix the problem with cdo though...

ocefpaf:
Looks like `nco` is suffering from the same problem as #780. PR inbound. I am working on a solution for `cdo` too. Looks like I'll have to package `libuuid`.

I wonder if changing from the CIs gcc to conda's gcc is the root of the problem...

ocefpaf:
Fixed in #797 and #796. Let's re-open if the problem re-surfaces, but the CIs are green.

PR #797, opened 2016-03-03
Fix cdo #776

ocefpaf:
Ping @DamienIrving.  Can you test this once the binaries land on the channel?

### DIFF
--- a/cdo/meta.yaml
+++ b/cdo/meta.yaml
@@ -8,7 +8,7 @@ source:
     md5: cca30c3c79335ad734e1838806f7bfc2
 
 build:
-    number: 1
+    number: 0
     skip: True  # [win]
 
 requirements:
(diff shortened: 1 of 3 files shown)


DamienIrving:
@ocefpaf No problem. I'll know when they've landed by looking [here](https://anaconda.org/IOOS/cdo/files), right?

ocefpaf:
>@ocefpaf No problem.

Thanks! I don't own a Mac to test.

>  I'll know when they've landed by looking here, right?

Yep.  The OS X CI (Travis-CI) says everything is OK. So merging it now. It should take a few hours for them to show up in the channel.

DamienIrving:
@ocefpaf As far as I can tell they still haven't landed on the channel. Is that normal?

cdo/meta.yaml
```diff
@@ -8,7 +8,7 @@ source:
     md5: cca30c3c79335ad734e1838806f7bfc2
 
 build:
-    number: 1
+    number: 0
```
ocefpaf: > @ocefpaf As far as I can tell they still haven't landed on the channel. Is that normal?

@DamienIrving my mistake.  This should be `2` and not `0`. Fixing this....

DamienIrving:
@ocefpaf Just tested on Mac - it's all working fine.

ocefpaf:
Thanks for testing it.

PR #803, opened 2016-03-15
Fetch from conda-forge

ocefpaf:
At some point we will either copy the binaries or clone the [feedstocks repo](https://github.com/conda-forge/feedstocks). For now I am doing this just to keep track of what was already sent to conda-forge.


pelson:
Sounds perfectly sensible to me. :+1:

PR #816, opened 2016-04-14
Conda build all

ocefpaf:
With `obvious-CI` we were  stuck with `conda-build 1.18.2`, by using `conda-build-all` we can updated to the latest version.


pelson:
:+1: on doing it. There is an unexpectedly large amount of noise here though. Intentional?

ocefpaf:
> :+1: on doing it. There is an unexpectedly large amount of noise here though. Intentional?

Yes. And there will be more noise. (I had a request to leave everything green and in working conditions in case we need to go back.)

PR #830, opened 2016-04-28
Latest geopandas

ocefpaf:
Ping @emiliom. The `conda-forge` version must be the latest release but we can relax that rule here. Let me know if that works for you.

### DIFF
--- a/recipes/geopandas/bld.bat
+++ b/recipes/geopandas/bld.bat
@@ -1,2 +1,0 @@
-"%PYTHON%" setup.py install --single-version-externally-managed --record record.txt
-if errorlevel 1 exit 1
--- a/recipes/geopandas/meta.yaml
+++ b/recipes/geopandas/meta.yaml
@@ -1,15 +1,18 @@
+{% set version = "0.2.0.dev0" %}
+{% set commit = "c15c2b7ebc7151261f00f1d29a838bfd82c6341e" %}
+
 package:
     name: geopandas
-    version: "0.1.1"
+    version: {{ version }}
 
 source:
-    fn: geopandas-0.1.1.tar.gz
-    url: https://pypi.python.org/packages/source/g/geopandas/geopandas-0.1.1.tar.gz
-    md5: a01b5d3eb6bc097665a0ba297c39029f
+    fn: geopandas-{{ version }}.tar.gz
+    url: https://github.com/geopandas/geopandas/archive/{{ commit }}.tar.gz
+    sha256: 0beb969e0a17a41076436ba7d536a5e763de9733665c42c6f902dc8490a5eaf0
 
 build:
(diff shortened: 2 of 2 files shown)


emiliom:
You mean using github master branch here, right? Yes, I think that's a good idea! The geopandas master branch has been fairly stable for a long time, I think. What's being worked on before a 0.2 release is mainly documentation (it's very outdated), I think; and small loose ends.

I actually was about to ask you about geopandas on the ODM2/BiGCZ conda recipes. But I think it's best to just update it here (IOOS recipes & channel) based on github master, then pick it up from conda-forge once an official 0.2 release is made.

Thanks!

ocefpaf:
> You mean using github master branch here, right?

Yes. I pinned to the latest commit (which is latest master today) so we can have reproducible builds. If I use `master` you never know exactly what went there unless you have the precise time of the `git clone` used in the build.

> I actually was about to ask you about geopandas on the ODM2/BiGCZ conda recipes. But I think it's best to just update it here (IOOS recipes & channel) based on github master, then pick it up from conda-forge once an official 0.2 release is made.

Agreed.

emiliom:
>  I pinned to the latest commit (which is latest master today) so we can have reproducible builds. If I use master you never know exactly what went there unless you have the precise time of the git clone used in the build.

Ah, yes, that makes perfect sense. Learning something every day.

emiliom:
FYI, at https://anaconda.org/IOOS/geopandas it looks like there's no build for OSX. Same for https://anaconda.org/IOOS/fiona (I assume Fiona is a geopandas dependency).

ocefpaf:
They are being built. See https://travis-ci.org/ioos/conda-recipes for the progress.

The problem is that Travis-CI, used for the OS X builds, limits the log to 4 MB and `gdal` exceeds that. So I need to hit re-start after each build.

emiliom:
> The problem is that Travis-CI, used for the OS X builds, limits the log to 4 MB and gdal exceeds that. So I need to hit re-start after each build.

Is there a paid subscription/option at Travis-CI that would increase the log limit and make the gdal build more automated (and less painful!)?

ocefpaf:
> Is there a paid subscription/option at Travis-CI that would increase the log limit and make the gdal build more automated (and less painful!)?

I am not sure if the log limitation is improved in the paid version, but that is not too painful (and there workarounds that I was too lazy to implement.) It happens only with `gdal`. In conda-forge we build way more complex stuff using the free account without a problem.

emiliom:
> I am not sure if the log limitation is improved in the paid version, but that is not too painful (and there workarounds that I was too lazy to implement.) It happens only with gdal. In conda-forge we build way more complex stuff using the free account without a problem.

Good. I won't get out my wallet yet :relieved:

PR #836, opened 2016-05-02
Cmake

ocefpaf:
Some fine tuning.


ocefpaf:
The number of time I have to merge a failing PR is the proof that the distribute model of conda-forge is better than the multiple recipes in 1 repo.

PR #842, opened 2016-05-06
Fix utilities build

ocefpaf:
Fixing #840

### DIFF
--- a/recipes/utilities/meta.yaml
+++ b/recipes/utilities/meta.yaml
@@ -15,12 +15,13 @@ build:
 
 requirements:
     build:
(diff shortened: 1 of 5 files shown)


recipes/utilities/meta.yaml
```diff
@@ -15,12 +15,13 @@ build:
 
 requirements:
     build:
+        - python
```
ocefpaf: @rsignell-usgs this was the issue! I somehow forgot the python dependency :flushed:, but I am taking the opportunity to fix some other builds too.

